PR #13, opened 2019-07-09
fix: append unprocessable errors to exception message

maxchehab:
* fix tests to use snapshot
* add snapshot

### DIFF
--- a/src/__snapshots__/workos.spec.ts.snap
+++ b/src/__snapshots__/workos.spec.ts.snap
@@ -4,7 +4,7 @@ exports[`WorkOS createEvent when the api responds with a 422 throws an Unprocess
 UnprocessableEntityException {
   "message": "The following requirements must be met:
 	target_id must be a string
-	occurred_at must be a ISOString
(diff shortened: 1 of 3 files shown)


src/__snapshots__/workos.spec.ts.snap
```diff
@@ -0,0 +1,12 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`WorkOS createEvent when the api responds with a 422 throws an UnprocessableEntity 1`] = `
+UnprocessableEntityException {
+  "message": "The following requirements must be met:
+	target_id must be a string
+	occurred_at must be a ISOString
```
marktran: Where does the copy for this error come from? Can we make it *occurred_at must be a ISO 8601 date string*?

*ISO* by itself just means *International Organization for Standardization* and reads ambiguously

maxchehab: It's from the api, I'll change it in https://github.com/workos-inc/workos/pull/452

PR #27, opened 2019-10-15
Maxchehab/ch767/refactor node sdk

maxchehab:
This PR implements changes described in [this story](https://app.clubhouse.io/workos/story/767/refactor-node-sdk-to-have-separate-submodules). Furthermore it:
- moves the `createEvent` function to a separate `auditLog` module.
- refactors constructor from `new WorkOS({ key: "sk_..." })` to `new WorkOS("sk_...")`
- refactors tests
- updates readme to reflect changes

### DIFF
--- a/src/audit-log/audit-log.spec.ts
+++ b/src/audit-log/audit-log.spec.ts
@@ -18,7 +18,7 @@ const event = {
 };
 
 describe('AuditLog', () => {
-  describe('create', () => {
+  describe('createEvent', () => {
     describe('when the api responds with a 201 CREATED', () => {
       it('posts Event successfuly', async () => {
         mock.onPost().reply(201, { success: true });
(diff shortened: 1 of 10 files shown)


src/audit-log/audit-log.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  target_name: 'Security Audit 2018',
+  target_id: 'document_39127',
+  action_type: 'U',
+  action: 'document.updated',
+};
+
+describe('AuditLog', () => {
+  describe('createEvent', () => {
+    describe('when the api responds with a 201 CREATED', () => {
+      it('posts Event successfuly', async () => {
```
sbauch: more a question than anything, but im not sure i follow what this is actually testing?

my naive reading is that we mock a 201, and then assert that there were no exceptions - in what world could this test fail?

maxchehab: Ah good question. This test is in place for any future commits that may happen to break the really simple functionality.

Although it seems unlikely that this test would fail, I generally add something akin to this for each spec.

PR #29, opened 2019-10-16
Maxchehab/ch768/add authorizeurl function to sso module

maxchehab:
This PR:
- Adds an SSO module to the SDK
	- It can be accessed by `new WorkOS().sso`
- Adds an `sso.getAuthorizationURL` which interpolates provided, `domain`, `projectID`, and `redirectURI`, along with using the optional `apiHostname` configuration, that generate a url to initiates our Oauth2 flow.
- Converts `WorkOS.key` and `WorkOS.options` from `private readonly` to just `readonly` so that submodules can reference their properties.

### DIFF
--- a/src/sso/sso.ts
+++ b/src/sso/sso.ts
@@ -7,7 +7,7 @@ import WorkOS from '../workos';
 export class SSO {
   constructor(private readonly workos: WorkOS) {}
 
-  authorizeURL({
+  getAuthorizationURL({
     domain,
     projectID,
     redirectURI,
--- a/src/sso/sso.ts
+++ b/src/sso/sso.ts
@@ -1,5 +1,27 @@
+import queryString from 'query-string';
+
+import { API_HOSTNAME } from '../common/constants';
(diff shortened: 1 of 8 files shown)


src/sso/sso.ts
```diff
@@ -0,0 +1,27 @@
+import queryString from 'query-string';
+
+import { API_HOSTNAME } from '../common/constants';
+import { AuthorizeURLOptions } from './interfaces/authorize-url-options.interface';
+import WorkOS from '../workos';
+
+export class SSO {
+  constructor(private readonly workos: WorkOS) {}
+
+  getAuthorizeURL({
```
marktran: `getAuthorizationURL`?

sbauch: I think I agree with Mark - "Authorization URL" feels like a standard term noun to me?

maxchehab: Sweet, just updated it @marktran @sbauch

PR #31, opened 2019-10-17
Maxchehab/ch775/add getprofile function to sso module

maxchehab:
This PR adds a `getProfile()` function.

It also addresses various typo's and bug where `getAuthorizationCode` also needed to send an additional query parameter (`response_type: 'code'`).

Example file: https://gist.github.com/maxchehab/c8569edb6e0958d3cf0960742ebb9895

### DIFF
--- a/src/sso/sso.ts
+++ b/src/sso/sso.ts
@@ -7,7 +7,7 @@ import WorkOS from '../workos';
 export class SSO {
   constructor(private readonly workos: WorkOS) {}
 
-  authorizeURL({
+  getAuthorizationURL({
     domain,
     projectID,
     redirectURI,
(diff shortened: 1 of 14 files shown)


src/sso/sso.ts
```diff
@@ -12,16 +14,33 @@ export class SSO {
...
     state,
-  }: AuthorizeURLOptions): string {
+  }: AuthorizationURLOptions): string {
     const { apiHostname = API_HOSTNAME } = this.workos.options;
 
     const query = queryString.stringify({
       domain,
       client_id: projectID,
       redirect_uri: redirectURI,
+      response_type: 'code',
```
jthodge: Just to make sure I'm understanding this correctly: the `response_type` value informs the queried auth server which type of grant (in this case, an auth code) to return in its response, correct?

maxchehab: yup!

jthodge: Great, thanks! 😄

PR #49, opened 2019-12-27
Maxchehab/ch1166/remove redirect uri from the getprofile function

maxchehab:
This PR:
- removes the `redirectURI` parameter from the `sso.getProfile` function and removes the `redirect_uri` query parameter from the `POST /sso/token` request.
- adds a [**Release Notes**](https://github.com/workos-inc/workos-js/tree/maxchehab/ch1166/remove-redirect-uri-from-the-getprofile-function#release-notes) section to the readme with migration examples.


README.md
```diff
@@ -119,3 +119,28 @@ Resulting in the following being sent to WorkOS:
 ```
 
 By adding supportive metadata when you create the event you can see what the original tweet body was and what the body was updated to. For something like a tweet that could get updated multiple times over the course of time, you can't always depend on the database representation to tell you what the body has always been. Without logging it right when the change occurs, you'll forever lose all the individual changes along the way. Good Audit Log events attach all supporting information surrounding the event which could be used to inform the reader in the future what exactly happened, how it happened, and when it happened.
+
+# Release Notes
```
marktran: How about a CHANGELOG file instead of putting this in the README?

Here's an example: https://github.com/typeorm/typeorm/blob/master/CHANGELOG.md

PR #58, opened 2020-01-13
Maxchehab/ch1326/node implement idempotency key header for

maxchehab:
This PR lets the developer attach a custom `Idempotency-Key` header to `POST /events`.

This feature is exposed by an optional parameter in the `createEvent` method. Ex: 
```ts
workos.auditLog.createEvent(event, {
  idempotencyKey: 'an-idempotency-key',
}),
```

### DIFF
--- a/src/audit-log/audit-log.ts
+++ b/src/audit-log/audit-log.ts
@@ -6,12 +6,6 @@ export class AuditLog {
   constructor(private readonly workos: WorkOS) {}
 
   async createEvent(event: Event, { idempotencyKey }: CreateEventOptions = {}) {
-    const headers: any = {};
-
-    if (idempotencyKey) {
-      headers['Idempotency-Key'] = idempotencyKey;
-    }
-
-    await this.workos.post('/events', event, { headers });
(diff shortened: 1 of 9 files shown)


marktran:
Is this ready for review @maxchehab ?

maxchehab:
@marktran yep

src/audit-log/audit-log.ts
```diff
@@ -1,10 +1,17 @@
...
 import { Event } from '../common/interfaces';
 import WorkOS from '../workos';
 
 export class AuditLog {
   constructor(private readonly workos: WorkOS) {}
 
-  async createEvent(event: Event) {
-    await this.workos.post('/events', event);
+  async createEvent(event: Event, { idempotencyKey }: CreateEventOptions = {}) {
+    const headers: any = {};
```
zfoster: worth having a header type? could even just specify known headers since we're building them here

```
const headers: { Idempotency-Key?: string } = {};
```

maxchehab: I like this. It's more strict. Will implement.

1.

PR #59, opened 2020-01-14
Allow `provider` param in AuthorizationURLOptions interface

rjvani:


src/sso/sso.ts
```diff
@@ -11,11 +11,19 @@ export class SSO {
   getAuthorizationURL({
     domain,
     projectID,
+    provider,
     redirectURI,
     state,
   }: AuthorizationURLOptions): string {
+    if (!domain && !provider) {
+      throw new Error(
+        `Incomplete arguments. Need to specify at least one of 'domain' or 'provider'.`,
```
maxchehab:
```suggestion
        `Incomplete arguments. Need to specify either a 'domain' or 'provider'.`,
```

rjvani: Fixed 👍

PR #90, opened 2020-03-13
refactor Semaphore

marktran:
Run commands CI commands in parallel

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -17,7 +17,7 @@ global_job_config:
     commands:
       - checkout
       - nvm use
-      - bash ~/bin/restore-or-install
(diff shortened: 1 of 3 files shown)


.semaphore/semaphore.yml
```diff
@@ -2,89 +2,47 @@ version: v1.0
...
 blocks:
   - name: Install dependencies
+    dependencies: []
     task:
-      secrets:
-        - name: workos-npm
-      env_vars:
-        - name: NODE_ENV
-          value: test
-        - name: CI
```
rjvani: Unused?

marktran: Yup, and we have `SEMAPHORE` available, which I think we should use instead.

PR #91, opened 2020-03-13
fix npm token

marktran:
Fix token in .npmrc

### DIFF
--- a/.semaphore/npm.yml
+++ b/.semaphore/npm.yml
@@ -11,9 +11,13 @@ blocks:
     task:
       secrets:
         - name: workos-npm
+      prologue:
+        commands:
+          - checkout
+          - nvm use
+          - bash ~/workos-node/bin/restore-or-install
       jobs:
         - name: Publish to NPM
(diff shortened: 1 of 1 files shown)


.semaphore/npm.yml
```diff
@@ -15,8 +15,9 @@ blocks:
         commands:
           - checkout
           - nvm use
-          - yarn install
+          - bash ~/workos-node/bin/restore-or-install
       jobs:
         - name: yarn publish
           commands:
+            - echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
             - yarn publish --access=public
```
maxchehab: Who's token is in there?

marktran: Yours?

maxchehab: Lets see what happens, but I think we need to create an account without 2fa, or without 2fa for publishing enabled.

Pretty sure npm will request the OTP from the stdin while the job runs.

PR #93, opened 2020-03-13
 Add authorization URL option for directory sync

rjvani:
Enabling flag to trigger directory sync workflow.


zfoster:
I was thinking we should have a new namespace for directory and keep `isDirectorySync` as an internal thing.

Like:

```
const url = workos.directory.getAuthorizationURL({
            domain: 'lyft.com',
            projectID: 'proj_123',
            redirectURI: 'example.com/sso/workos/callback',
          });
```

and it calls the same stuff under the hood but with the flag

rjvani:
@zfoster - I like that, updated.

PR #95, opened 2020-03-15
[Chore] rename audit log to audit trail

maxchehab:
Renames audit log to audit trail.

To upgrade:

```ts
// Current
const client = new WorkOS();
client.auditLog.createEvent(...);

// New
const client = new WorkOS();
client.auditTrail.createEvent(...);
```


marktran:
Let's bump all the SDKs and jump to 0.2.0

PR #96, opened 2020-03-15
[Fix] update package version

maxchehab:
Accidentally published an incorrect version under `0.2.0`. I ran `npm unpublish` to revert, but in order to publish the correct version we need to bump a version.


maxchehab:
@marktran I removed the incorrect version from the registry, however they assert that you can't rewrite it.

https://docs.npmjs.com/cli/unpublish
>Even if a package version is unpublished, that specific name and version combination can never be reused. In order to publish the package again, a new version number must be used. Additionally, new versions of packages with every version unpublished may not be republished until 24 hours have passed.

PR #125, opened 2020-04-16
Create directory sync module

maxchehab:
This PR creates a Directory Sync module where one can:
- List Directories
  ```ts
  client.directorySync.listDirectories()
  ```
- List Users
  ```ts
  client.directorySync.listUsers("dir_edp_123")
  ```
- List Groups
  ```ts
  client.directorySync.listGroups("dir_edp_123")
  ```
- Get a User
  ```ts
  client.directorySync.getUser("dir_edp_123", "dir_usr_123")
  ```
- List User's Groups
  ```ts
  client.directorySync.listUsersGroups("dir_edp_123", "dir_usr_123")
  ```


rjvani:
I think we need to add fetching a single user: https://docs.workos.com/directory-sync/api-reference#retrieve-a-single-user.

maxchehab:
> I think we need to add fetching a single user: https://docs.workos.com/directory-sync/api-reference#retrieve-a-single-user.

Done

src/directory-sync/directory-sync.ts
```diff
@@ -0,0 +1,54 @@
...
+    options?: PaginationOptions,
+  ): Promise<List<User>> {
+    const { data } = await this.workos.get(
+      `/directories/${directoryID}/users`,
+      options,
+    );
+    return data;
+  }
+
+  async listUsersGroups(directoryID: string, userID: string): Promise<Group[]> {
```
marktran, line 41: Should this be `listUserGroups`?

We're grabbing groups for a single user, right?

maxchehab: Shouldn't the noun be possessive?

marktran: I always think about it as singular/plural. Don't think I've ever seen a possessive in API naming.

maxchehab: 👍 https://github.com/workos-inc/workos-node/pull/126

PR #126, opened 2020-04-17
Remove getUserGroup posesive

maxchehab:


maxchehab:
@marktran They both follow the non-possessive version.

PR #138, opened 2020-04-30
Feat/add connection endpoint for promoting draft connection

maxchehab:
Creates a `sso.createConnection({ source })` method. Should be used in replacement of `sso.promoteDraftConnection({ token })`.

A depreciation message was added to `sso.promoteDraftConnection`. We should remove the method completely at the next major version.

### DIFF
--- a/src/sso/sso.spec.ts
+++ b/src/sso/sso.spec.ts
@@ -154,5 +154,45 @@ describe('SSO', () => {
         }
       });
     });
+
+    describe('createConnection', () => {
+      it('send valid token', async () => {
+        const source = 'wOrkOStoKeN';
+
+        const mock = new MockAdapter(axios);
+        mock.onPost('/connections').reply(201);
+
+        const workos = new WorkOS('sk_test_Sz3IQjepeSWaI4cMS4ms4sMuU');
+        await workos.sso.createConnection({ source });
+
+        expect(mock.history.post.length).toBe(1);
+        const post = mock.history.post[0];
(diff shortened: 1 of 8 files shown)


src/sso/sso.spec.ts
```diff
@@ -154,5 +154,45 @@ describe('SSO', () => {
         }
       });
     });
+
+    describe('createConnection', () => {
+      it('send valid token', async () => {
```
marktran, line 159:
```
describe('with valid token', () => {
  it('responds with a 201 CREATED status', () => {
  })
})

describe('with invalid token', () => {
  it('responds with a 404 NOT FOUND status', () => {
  })
})
```

src/sso/sso.ts
```diff
@@ -46,6 +55,11 @@ export class SSO {
   }
 
   async promoteDraftConnection({ token }: PromoteDraftConnectionOptions) {
+    //  tslint:disable:no-console
+    console.error(
```
marktran, line 59: Can we add a logger instead of using console.*

maxchehab:
Added an optional `logger: stream.Transform` which defaults to a Winston instance. Developers can provide their own loggers as an optional parameter.

Update `createConnection` and `promoteDraftConnection` tests.

src/sso/sso.ts
marktran: How about `console.warn`? Seems more appropriate for a deprecation.

PR #197, opened 2020-08-06
Change Audit Trail event type to string

rjvani:
Also adding a barrel files for exports.


marktran:
@rjvani What does it look like on the other side?

src/sso/sso.ts
```diff
@@ -6,7 +6,7 @@ import { CreateConnectionOptions } from './interfaces/create-connection-options.
 import { GetProfileOptions } from './interfaces/get-profile-options.interface';
 import { Profile } from './interfaces/profile.interface';
 import { PromoteDraftConnectionOptions } from './interfaces/promote-draft-connection-options.interface';
-import WorkOS from '../workos';
+import WorkOS from '../';
```
marktran, line 9: This seems weird, I think we should just explicitly `from '../workos'`

PR #217, opened 2020-09-10
Add Portal and expose listOrganizations()

willmanduffy:
* Returns a list of the organizations accessible to the provided
API key. Includes pagination functionality via the listMetadata
property.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "build": "tsc -p tsconfig.json",
     "lint": "tslint -p tsconfig.json -c tslint.json",
     "test": "jest",
+    "test:watch": "jest --watch",
     "prettier": "prettier \"src/**/*.{js,ts,tsx}\" --check",
     "format": "prettier \"src/**/*.{js,ts,tsx}\" --write",
     "prepublishOnly": "yarn run build"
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -27,6 +27,7 @@
     "build": "tsc -p tsconfig.json",
     "lint": "tslint -p tsconfig.json -c tslint.json",
     "test": "jest",
+    "test:watch": "jest --watch",
```
willmanduffy, line 30: Just adding this as a convenience for myself to allow running tests on each file change.

src/portal/fixtures/list-organizations.json
```diff
@@ -0,0 +1,90 @@
+{
```
willmanduffy, line 1: We haven't done this pattern yet! I wanted to test what the SDK does with the data it receives and confirm that it drops it in an acceptable data structure. Previous similar tests were just testing that we formed the correct request to the API.

It's super easy to generate these responses in Postman or, in this case, I just grabbed it from the Ruby SDK VCR cassettes.

maxchehab: I like it 👍

maxchehab (approved):
🚢

src/portal/fixtures/list-organizations.json
maxchehab: I like it 👍

PR #223, opened 2020-09-14
@willmanduffy Bump @workos-inc/node to 0.6.1

willmanduffy:
* Had an error with the release of 0.6.0 and the `package.json` did not properly update. Bumping us to 0.6.1 to get us back into a proper state.


maxchehab (approved):
lgtm

PR #224, opened 2020-09-15
Add MagicLink support through Passwordless module

zlwaterfield:
Add Passwordless support for MagicLink. Adds two methods
 - `createSession` - generate a passwordless session for a user from their email
 - `sendSession` - send an email to a user with a Magic Link confirmation URL


src/workos.ts
```diff
@@ -23,6 +24,7 @@ export class WorkOS {
   readonly auditTrail = new AuditTrail(this);
   readonly directorySync = new DirectorySync(this);
   readonly portal = new Portal(this);
+  readonly passwordless = new Passwordless(this);
```
willmanduffy, line 27: Nit: Looks like the rest of this list is alphabetized so it might be nice to bump it up a line.

src/passwordless/passwordless.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const mock = new MockAdapater(axios);
+
+describe('Passwordless', () => {
+  afterEach(() => mock.resetHistory());
+
+  describe('createSession', () => {
+    describe('with valid options', () => {
+      it('creates a passwordless session', async () => {
+        mock.onPost().reply(200, {});
```
willmanduffy, line 14: Does this return a 200 or a 201?

willmanduffy: Recently I started using a fixture pattern of responses so we not only test that the SDK forms the proper request but we also test that it handles the response properly. You can see an example here. Might be nice to throw in!

https://github.com/workos-inc/workos-node/pull/218/files#diff-23bdba8287c2e5e301c52a9d04fe0c46R19

willmanduffy (approved):
Looks good! Just a couple thoughts to peruse

src/workos.ts
zlwaterfield: - [ ] fix module ordering

src/passwordless/passwordless.spec.ts
zlwaterfield: Just checked and it is a 201, I can update it. And thanks for the share, I'll update the test to follow that pattern.

zlwaterfield: - [ ] update test to use fixtures
- [ ] update test response to be a 201

zlwaterfield:
Tested locally using a sandbox API key and it is working.

PR #241, opened 2020-10-13
Expose `connection_id` on Profile Objects

jthodge:
This PR introduces the following changes:

* Passes the `connection_id` Profile object attribute through to the SDK. We currently expose `connection_id` via the `/sso/token` API endpoint.
* Updates the SDK version to `0.8.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.7.0",
+  "version": "0.8.0",
   "name": "@workos-inc/node",
   "author": "WorkOS",
   "description": "A Node wrapper for the WorkOS API",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.7.0",
+  "version": "0.8.0",
```
maxchehab, line 2: Lets just bump the minor version.

maxchehab (approved):

package.json
jthodge: @maxchehab I think this would be minor bump based on our [versioning policy](https://github.com/workos-inc/workos/wiki/SDK-Release-Process#versioning-policy) since this isn’t a bug fix.

PR #265, opened 2020-11-17
Add `CODEOWNERS`

jthodge:
This PR adds a `CODEOWNERS` file to automatically request reviews from the associated and responsible WorkOS team members.


bukinoshita (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# See GitHub's docs for more details:
+# https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
+
+# TypeScript Team
+* @workos-inc/typescript
```
zlwaterfield, line 5: This is is public repo but I don't think our teams are public: https://github.com/orgs/workos-inc/teams/typescript

maxdeviant: Do the teams need to be public for `CODEOWNERS` to work?

marktran: I don't believe it does. Seems to register correctly:
> bukinoshita approved these changes on behalf of workos-inc/typescript 6 minutes ago

maxdeviant: Yea, I think this is fine.

GitHub is still aware of the private team even if you aren't part of the @workos-inc org (this is how it looks if you're not signed into GitHub):

Screen Shot 2020-11-17 at 3 47 51 PM

maxdeviant (approved):

.github/CODEOWNERS
jthodge: Yep! Thanks for the questions and responses, y'all. Going to merge :shipit:

PR #273, opened 2020-11-23
Use the correct parameter name for `redirect_uri`

maxdeviant:
This PR fixes an issue with creating a passwordless session where we were using the wrong name when passing the `redirect_uri` up to the API.


jthodge (approved):
❤️

PR #274, opened 2020-11-24
Add a changelog

maxdeviant:
This PR adds a [changelog](https://keepachangelog.com/en/1.0.0/) to keep track of changes within the Node SDK.

The changelog has been backfilled with the release notes from each version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,140 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [0.8.3] - 2020-11-23
+
+### Fixed
+
+- Fixed an issue where the `redirectURI` parameter was being passed up to the API using the wrong name ([#273](https://github.com/workos-inc/workos-node/pull/273))
+
+## [0.8.2] - 2020-11-19
+
+### Changed
+
+- Improved naming consistency for better ergonomics ([#269](https://github.com/workos-inc/workos-node/pull/269))
+
+## [0.8.1] - 2020-11-18
+
+### Added
+
+- Support `redirect_uri` Option for Creating Passwordless Sessions ([#267](https://github.com/workos-inc/workos-node/pull/267))
+
+## [0.8.0] - 2020-10-14
+
+### Added
+
+- Exposed `connection_id` attribute on Profile objects. ([#241](https://github.com/workos-inc/workos-node/pull/241))
+
+## [0.7.0] - 2020-09-16
+
+### Added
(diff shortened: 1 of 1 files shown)


zlwaterfield (approved):
Is this auto-generated or is this done manually for each release?

maxdeviant:
> Is this auto-generated or is this done manually for each release?

It's currently manual.

The idea is that while changes are made they get added to the `Unreleased` section and then when we cut a release we just rename `Unreleased` to the new version.

marktran (changes requested):
What's the advantage of a `CHANGELOG` file vs GitHub Releases? What's the process for updating it and making releases?

If we add it to this SDK, it should also be added to all the other SDKs for consistency.

maxdeviant:
> What's the advantage of a `CHANGELOG` file vs GitHub Releases?

The main advantage of having a `CHANGELOG` is that you can have a draft state of all unreleased changes. So every time a change is made, you can add an entry under `Unreleased`. Then when you cut a release, you don't need to go back and try to remember all of the changes that are going into the release.

Additionally, the `CHANGELOG` in the repo is versioned with the project, and is more easily accessible to library consumers, especially if they want to view the changes since some very old version (as they don't have to dig through pages of GitHub releases).

The two are not mutually exclusive, and what projects will typically do is create a GitHub Release and take the contents of the changelog for that version and copy-paste it into the GitHub Release description.

Here's what the Keep a Changelog site says on the matter:

> It's a great initiative. Releases can be used to turn simple git tags (for example a tag named v1.0.0) into rich release notes by manually adding release notes or it can pull annotated git tag messages and turn them into notes.
> 
> GitHub Releases create a non-portable changelog that can only be displayed to users within the context of GitHub. It's possible to make them look very much like the Keep a Changelog format, but it tends to be a bit more involved.
>
> The current version of GitHub releases is also arguably not very discoverable by end-users, unlike the typical uppercase files (README, CONTRIBUTING, etc.). Another minor issue is that the interface doesn't currently offer links to commit logs between each release.

---

> What's the process for updating it and making releases?

Each time a change is made it is added under the `Unreleased` section in `CHANGELOG.md` under the corresponding section. Keep a Changelog defines the following sections:

- `Added` for new features.
- `Changed` for changes in existing functionality.
- `Deprecated` for soon-to-be removed features.
- `Removed` for now removed features.
- `Fixed` for any bug fixes.
- `Security` in case of vulnerabilities.

For example, a PR that fixes a bug would add a line about the fix under `Fixed`.

When making a release you would change `Unreleased` to the version being released, add a date, update the links at bottom of the changelog, and add a new, empty `Unreleased` section at the top. If the changes were kept on top of as they were made then the maintainer cutting the release shouldn't need to make any changes to the content of the changelog for that release.

> If we add it to this SDK, it should also be added to all the other SDKs for consistency.

I agree that we would want to adopt this across the board.

maxdeviant:
Closing this out until we decide to revisit.

PR #284, opened 2020-12-18
Removing project_id from directory and event interfaces

logangingerich:
https://app.clubhouse.io/workos/story/4514/node-remove-project-id-from-directory-and-event-interfaces

Removes `project_id` from the directory and event interfaces

PR #285, opened 2020-12-21
Add `client_id` to SSO package

loferris:
Story details: https://app.clubhouse.io/workos/story/4278


maxdeviant (dismissed):

maxdeviant:
@loferris Could you remove the `package-lock.json` since we use Yarn?

package-lock.json
```diff
@@ -0,0 +1,5149 @@
+{
```
marktran, line 1: Can you please remove this file

src/sso/interfaces/authorization-url-options.interface.ts
```diff
@@ -1,6 +1,6 @@
 export interface AuthorizationURLOptions {
   domain?: string;
-  projectID: string;
+  clientID: string;
```
marktran, line 3: Looks like the Clubhouse story is for adding `clientID` and keeping `projectID` (to maintain backward compatibility)?

marktran (changes requested):

src/sso/interfaces/authorization-url-options.interface.ts
loferris: I added projectID back as a supported paramater and rewrote the type interfaces and the tests to reflect in my most recent commit @marktran

loferris: second commit resolved failing semaphore check

src/sso/interfaces/authorization-url-options.interface.ts
```diff
@@ -1,6 +1,7 @@
 export interface AuthorizationURLOptions {
   domain?: string;
-  projectID: string;
+  projectID?: string | undefined;
```
maxdeviant, line 3: We don't actually need `| undefined` here, as making a property as optional (`?`) already widens the type to include `undefined`.

src/sso/sso.spec.ts
```diff
@@ -6,7 +6,21 @@ import { WorkOS } from '../workos';
 describe('SSO', () => {
   describe('SSO', () => {
     describe('getAuthorizationURL', () => {
-      describe('with no custom api hostname', () => {
+      describe('with no custom api hostname and clientID', () => {
```
marktran, line 9: I don't think we need to explicitly add "and clientID" to the `describe` blocks in the tests in this file.

`clientID` will be a required parameter and these tests are meant to enumerate on optional parameters.

loferris: done! I also asked this question about the tests on Slack and got similar responses: https://work-os.slack.com/archives/CPEAGQU8G/p1608669941038900

src/sso/interfaces/authorization-url-options.interface.ts
loferris: done

src/sso/interfaces/authorization-url-options.interface.ts
maxdeviant: > Looks like the Clubhouse story is for adding `clientID` and keeping `projectID` (to maintain backward compatibility)?

@marktran Looking at this again, I think the intention is to replace `projectID` with `clientID`. The `client_id` column on a V2 Project corresponds to the `id` column on a V1 Project.

I believe we had said that we were going to make this a breaking change and bump the major versions of the SDKs, so that when consumers update they would replace `projectID` with `clientID`.

We're already making breaking changes by removing `project_id` and replacing it with `environment_id` (see #284 and #286).

marktran: @maxdeviant I thought the`client_id` column on V2 Project was backfilled with the `id` column from a V1 Project for existing users but had a generated id in a new format for new users?

Do you know if this migration strategy is documented anywhere? I was under the impression that we could make all of this backwards compatible so current users wouldn't have to change anything.

maxdeviant: @marktran I'm not familiar with the specifics of the `client_id` for new V2 projects.

I'm trying to find where we discussed the migration strategy for the SDKs. I definitely recall us saying that it would be a major version bump, which would indicate that we were making breaking changes. I have a feeling this was something that was said verbally and perhaps wasn't written down...

marktran: We should probably get clarification from @rjvani before proceeding on these SDK changes. These stories don't have enough context.

I think there are some implications to making this a breaking change:
- Friction for current users to upgrade.
- Some current users will hold off on upgrading and we'll be forced to maintain compatibility with old versions 
- We'll need to push people to upgrade and that'll ultimately slow us down

loferris: @marktran and @maxdeviant Should I withdraw this PR and just sit on the other client_id tickets until Rohan gets back? Or does this apply to all current SDK tickets that include breaking changes?

marktran: @loferris I think we can keep this open and hold off on Clubhouse stories that have potentially breaking changes.

loferris:
I have made changes reflecting the resolution @marktran and @rjvani clarified by Monday.

src/sso/sso.ts
```diff
@@ -43,9 +43,22 @@ export class SSO {
...
-    form.set('client_id', clientID);
+    if (clientID) {
+      form.set('client_id', clientID);
+    } else if (projectID) {
+      form.set('project_id', projectID);
+      this.workos.emitWarning(
+        `Use of project_id is deprecated; please use client_id in Dashboard2`,
+      );
+    } else {
+      this.workos.emitWarning(`You must enter a client_id or project_id`);
```
zlwaterfield, line 60: Should this return and throw an error here? Or not emit a warning and let the API throw an error? It feels weird to throw a warning when it should be an error.

loferris: @zlwaterfield anything else we shld do before merging?

jthodge:
@loferris Are we alright to close this PR?

loferris:
> @loferris Are we alright to close this PR?

I actually don't know - I've been asking for feedback/approval for a while. Did this become redundant because of one of Logan's PRs?

marktran:
> > @loferris Are we alright to close this PR?
> 
> I actually don't know - I've been asking for feedback/approval for a while. Did this become redundant because of one of Logan's PRs?

Yeah, this was covered by Logan's PR. Closing...

PR #293, opened 2021-01-05
Add client_id to AuthorizationUrlOptions and GetProfileOptions

logangingerich:
https://app.clubhouse.io/workos/story/4512/node-add-client-id-to-authorizationurloptions-and-getprofileoptions

Adds client_id to AuthorizationUrlOptions and GetProfileOptions


src/sso/sso.spec.ts
```diff
@@ -58,7 +61,7 @@ describe('SSO', () => {
 
           const url = workos.sso.getAuthorizationURL({
             domain: 'lyft.com',
-            projectID: 'proj_123',
+            clientID: 'proj_123',
```
jthodge, line 64: Should we be replacing the other pre-existing `projectID` parameters in this spec with `clientID` as well?

logangingerich: Removed projectID from the specs. Definitely let me know if anything looks funky still! 🙏

rjvani (approved):

PR #304, opened 2021-01-26
Allow clientID or projectID for backwards compatibility

logangingerich:
https://linear.app/workos/issue/SDK-26/allow-clientid-or-projectid-in-nodets-interfaces

Allows for backwards compatibility with projectID or clientID in Node/TS interfaces


maxdeviant:
We probably want to make this same change to anywhere that uses a `clientID` or `projectID` (e.g., `AuthorizationURLOptions`).

maxdeviant (approved):

logangingerich:
bors merge

PR #306, opened 2021-01-27
Add deleteConnection method

logangingerich:
https://linear.app/workos/issue/SDK-17/add-delete-connectionsid-for-node-sdk

Adds a `deleteConnection` method to wrap `DELETE /connections/:id` endpoint.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.8.4",
+  "version": "0.8.5",
   "name": "@workos-inc/node",
   "author": "WorkOS",
   "description": "A Node wrapper for the WorkOS API",
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "jest": "26.6.3",
     "prettier": "2.2.1",
     "supertest": "6.1.3",
-    "ts-jest": "26.4.4",
+    "ts-jest": "26.5.0",
     "tslint": "6.1.3",
     "typescript": "4.1.3"
   }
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.8.4",
+  "version": "0.8.5",
```
marktran, line 2: Can you break the version bump into a separate PR?

PRs get squashed into a single commit so we want the ability to easily revert functionality.

src/sso/sso.spec.ts
```diff
@@ -258,6 +258,18 @@ describe('SSO', () => {
       });
     });
 
+    describe('deleteConnection', () => {
+      it(`sends request to delete a Connection`, async () => {
```
marktran, line 262:
```suggestion
      it('sends request to delete a Connection', async () => {
```

package.json
logangingerich: @marktran Absolutely, I'll update shortly 👍

maxdeviant: Also, since we're adding functionality we should be bumping the minor version to follow [SemVer](https://semver.org/):
> 2. MINOR version when you add functionality in a backwards compatible manner, and

logangingerich: Added a commit reverting package.json version and snapshots back to 0.8.4

logangingerich: @maxdeviant I can plan for a bump to 0.9.0 (sorry, bit of a new area for me 😅 )

maxdeviant (approved):

PR #334, opened 2021-03-03
Add dsync intent for admin portal

blairlunceford:


src/portal/interfaces/generate-portal-link-intent.interface.ts
```diff
@@ -1,3 +1,4 @@
 export enum GeneratePortalLinkIntent {
   SSO = 'sso',
+  DSYNC = 'dsync',
```
willmanduffy, line 3: I actually have no idea if this should be `DSync` `Dsync` or `DSYNC`! I feel like `DSync` might be right?

maxdeviant: Yea, enum members should be named using `PascalCase`, so either `Dsync` or `DSync, depending on what sect of naming conventions you adhered to 😅

blairlunceford: Thanks! I updated it to `DSync` 👍

maxdeviant (approved):

PR #375, opened 2021-05-05
Update connection status to connection state

logangingerich:
https://linear.app/workos/issue/SDK-115/update-connection-status-to-connection-state-in-node-sdk

### DIFF
--- a/src/sso/interfaces/connection.interface.ts
+++ b/src/sso/interfaces/connection.interface.ts
@@ -9,7 +9,7 @@ export interface ConnectionDomain {
 export interface Connection {
   object: 'connection';
   id: string;
-  status: 'linked' | 'unlinked';
+  state: 'linked' | 'unlinked';
   name: string;
   connection_type: ConnectionType;
   oauth_uid: string;


bukinoshita (approved):

src/sso/interfaces/connection.interface.ts
```diff
@@ -9,7 +9,7 @@ export interface ConnectionDomain {
 export interface Connection {
   object: 'connection';
   id: string;
-  status: 'linked' | 'unlinked';
+  state: 'linked' | 'unlinked';
```
maxdeviant, line 12: @logangingerich The values here aren't correct. They should be `state: 'draft' | 'active' | 'inactive'`.

logangingerich: Ah, the Linear ticket didn't have that context. I'll update.

maxdeviant: We should also go through a migratory period where we support both in tandem temporarily, similar to what I mentioned here: https://github.com/workos-inc/workos-dotnet/pull/37#discussion_r626838549

PR #376, opened 2021-05-05
Feature/sdk 115 update connection status to connection

logangingerich:
Bumps minor version for breaking change, updates `state` values and adds back `status` for backwards compatability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.11.0",
+  "version": "0.10.2",
   "name": "@workos-inc/node",
   "author": "WorkOS",
   "description": "A Node wrapper for the WorkOS API",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.10.1",
+  "version": "0.11.0",
   "name": "@workos-inc/node",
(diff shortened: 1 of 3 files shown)


maxdeviant:
What breaking changes are in this release?

Since we added back `status`, the addition of `state` shouldn't be a breaking change.

logangingerich:
> What breaking changes are in this release?
> 
> Since we added back `status`, the addition of `state` shouldn't be a breaking change.

@maxdeviant I was following @blairlunceford 's note here https://linear.app/workos/issue/SDK-113#comment-73e166e1 but can revert

maxdeviant:
> > What breaking changes are in this release?
> > Since we added back `status`, the addition of `state` shouldn't be a breaking change.
> 
> @maxdeviant I was following @blairlunceford 's note here https://linear.app/workos/issue/SDK-113#comment-73e166e1 but can revert

Ah, gotcha. I think in this case since we can pretty easily avoid a breaking change by having both `status` and state` we should not make this a breaking change.

When we eventually remove the `status` field then we would bump the version to represent a breaking change.

maxdeviant (approved):

PR #394, opened 2021-05-12
Add Get an Organization Method

logangingerich:
https://linear.app/workos/issue/SDK-146/add-get-an-organization-method-to-node-sdk


src/organizations/organizations.spec.ts
```diff
@@ -141,6 +141,21 @@ describe('Organizations', () => {
     });
   });
 
+  describe('getOrganization', () => {
+    it(`requests an Organization`, async () => {
+      const mock = new MockAdapter(axios);
+      mock.onGet().reply(200, {});
```
maxdeviant, line 147: Can we mock out the response so we can test the value returned from `getOrganization`?

maxdeviant (approved):

PR #410, opened 2021-06-15
Improve error handling, Expose `error` and `error_description`

logangingerich:
https://linear.app/workos/issue/SDK-72/improve-error-handling-in-node-sdk

Exposes `error` and `error_description` for error handling.


src/common/exceptions/generic-server.exception.ts
```diff
@@ -6,10 +6,13 @@ export class GenericServerException extends Error {
...
     message: string | undefined,
     readonly requestID: string,
+    readonly err: string | undefined,
+    readonly errDescription: string | undefined,
   ) {
     super();
     if (message) {
-      this.message = message;
+      this.message =
+        message + `\nError: ${err}\nError Description: ${errDescription}`;
```
maxdeviant, line 15: We should probably only include the `err` and `errDescription` values if they have a value, because right now we have `undefined` coming through in the resulting error message.

src/common/exceptions/generic-server.exception.ts
```diff
@@ -6,10 +6,13 @@ export class GenericServerException extends Error {
     readonly status: number,
     message: string | undefined,
     readonly requestID: string,
+    readonly err: string | undefined,
```
maxdeviant, line 9: Can we expand these names to `error` and `errorDescription`?

src/common/exceptions/not-found.exception.ts
```diff
@@ -3,8 +3,13 @@ export class NotFoundException extends Error {
   readonly name: string = 'NotFoundException';
   readonly message: string;
 
-  constructor(path: string, readonly requestID: string) {
+  constructor(
+    path: string,
+    readonly requestID: string,
+    readonly err: string,
```
maxdeviant, line 9: Also, we should only get back `error` and `error_description` from 400 Bad Request errors, so we shouldn't add these properties to the other classes of errors.

maxdeviant (changes requested):
Instead of including the `error` and `error_description` values in the `GenericServiceException` class we should add a separate `OauthException` and use that error if we have either an `error` or an `error_description` on the response.

src/common/exceptions/oauth.exception.ts
```diff
@@ -0,0 +1,17 @@
...
+    readonly requestID: string,
+    readonly error: string | undefined,
+    readonly errorDescription: string | undefined,
+  ) {
+    super();
+    if (error && errorDescription) {
+      this.message = `Error: ${error}\nError Description: ${errorDescription}`;
+    } else if (error) {
+      this.message = `Error: ${error}`;
+    }
```
maxdeviant, line 15: We should have a default message in here in case both `error` and `errorDescription` are `undefined`.

src/workos.ts
```diff
@@ -100,7 +105,18 @@ export class WorkOS {
             throw new NotFoundException(path, requestID);
           }
           default: {
-            throw new GenericServerException(status, data.message, requestID);
+            if (error && errorDescription) {
```
maxdeviant, line 108: Since we allow for `error` and `error_description` to be `undefined`, we should be able to collapse two of these conditions into one:

```ts
if (error || errorDescription) {
  throw new OauthException(
    status,
    requestID,
    error,
    errorDescription
  );
} else {
  throw new GenericServerException(status, data.message, requestID);
}
```

src/workos.ts
```diff
@@ -127,6 +143,10 @@ export class WorkOS {
       if (response) {
         const { status, data, headers } = response;
         const requestID = headers['X-Request-ID'];
+        const err = 'error';
```
maxdeviant, line 146: Could we access this properties would storing the keys in new variables?

```ts
const error = data.error;
const description = data['error_description'];
```

We might even be able to write it like this:

```ts
const { error, error_description: errorDescription } = data;
```

src/workos.ts
maxdeviant: It looks like there's a couple of spots where we'll need to do this.

maxdeviant (approved):

PR #423, opened 2021-07-09
Add named export for CommonJS consumers

maxdeviant:
This PR adds a named `WorkOS` export for the SDK to facilitate more ergonomic usage by CommonJS users.

#### CommonJS

```js
const { WorkOS } = require('@workos-inc/node');

const workos = new WorkOS('sk_example_123456789');
```

#### ES Modules

```ts
import WorkOS from '@workos-inc/node';

const workos = new WorkOS('sk_example_123456789');
```

or

```ts
import { WorkOS } from '@workos-inc/node';

const workos = new WorkOS('sk_example_123456789');
```

Resolves SDK-76.

### DIFF
--- a/src/workos.ts
+++ b/src/workos.ts
@@ -1,5 +1,4 @@
 import axios, { AxiosError, AxiosInstance, AxiosResponse } from 'axios';
-import { version } from '../package.json';
 import { AuditTrail } from './audit-trail/audit-trail';
 import {
   GenericServerException,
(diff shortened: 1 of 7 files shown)


marktran (approved):

src/workos.ts
```diff
@@ -1,5 +1,4 @@
 import axios, { AxiosError, AxiosInstance, AxiosResponse } from 'axios';
-import { version } from '../package.json';
```
maxdeviant, line 2: We're not importing from `package.json` (which is outside of the `src` directory) so that we can flatten the output directory.

PR #504, opened 2021-10-29
Make `organization_id` nullable

maxdeviant:
This PR makes the `organization_id` field nullable on the following objects in preparation for future changes:

- `Connection`
- `Directory`
- `Profile`

This is a breaking change, so we'll need to bump the major version.

Resolves SDK-308.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.5.0",
+  "version": "1.6.0",
   "name": "@workos-inc/node",
   "author": "WorkOS",
   "description": "A Node wrapper for the WorkOS API",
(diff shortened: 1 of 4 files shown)


liaujianjie:
Under what circumstances would a connection not have an organization ID?

maxdeviant:
> Under what circumstances would a connection not have an organization ID?

Today all connections will have an organization ID.

We're currently investigating some changes to the WorkOS platform that will allow connections to exist outside of organizations. We don't have anything to announce on this just yet, but we'll be reaching out to impacted developers before making any changes would impact that value of the `organization_id` at runtime.

PR #512, opened 2021-11-05
Bump version to 2.1.0

logangingerich:


mthadley:
I think we'll want to recreate the `2.1.0` tag, so since it doesn't currently contain these version bumps.

So next time, let's create the release/tag after merging the version bump.

mthadley (approved):

logangingerich:
> I think we'll want to recreate the `2.1.0` tag, so since it doesn't currently contain these version bumps.
> 
> So next time, let's create the release/tag after merging the version bump.

@mthadley Ah thanks for the heads up, I’ll definitely recreate it!

PR #513, opened 2021-11-08
fix: add `directory_id` property to `DsyncUser*Webhook` interfaces

liaujianjie:
### Context

The `directory_id` property is observed in the webhook request body but it is not available on the `DsyncUser*Webhook` interfaces.

This PR adds the missing `directory_id` property to all directory sync webhook event interfaces' `data` property.

### Example `dsync.user.created` webhook request body

```js
{
    "id": "wh_01FKZ41WJYEG7ZQTAN6AZ47P6P",
    "data": {
        // ...
        "last_name": "Liau",
        "first_name": "Jian Jie",
        "directory_id": "directory_01FKZ3ZTD2Z5F1093YTKNBWVMJ", // <- this is present but not represented in the interface
        // ...
    },
    "event": "dsync.user.created"
}
```

### DIFF
--- a/src/webhooks/interfaces/webhook.interface.ts
+++ b/src/webhooks/interfaces/webhook.interface.ts
@@ -77,17 +77,23 @@ export interface DsyncGroupUserRemovedWebhook extends WebhookBase {
 
 export interface DsyncUserCreatedWebhook extends WebhookBase {
   event: 'dsync.user.created';
-  data: User;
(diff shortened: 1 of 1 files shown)


maxdeviant:
Thanks for the PR!

It looks like this change was already merged in #514.

We'll have a new version out soon with the changes 🙂

PR #514, opened 2021-11-08
Fix missing directory_id on user webhook event

andyrichardson:
# About

Fix missing attribute on user webhook events

> Note: I haven't looked into whether `directory_id` should be present on all uses of the `User` interface (or if this "merged" type is only used in webhooks

### DIFF
--- a/src/directory-sync/interfaces/group.interface.ts
+++ b/src/directory-sync/interfaces/group.interface.ts
@@ -1,5 +1,6 @@
 export interface Group {
   id: string;
+  directory_id: string;
   name: string;
   raw_attributes: any;
 }
--- a/src/directory-sync/interfaces/user.interface.ts
+++ b/src/directory-sync/interfaces/user.interface.ts
@@ -6,6 +6,7 @@ export interface User<
   TCustomAttributes extends object = DefaultCustomAttributes,
 > {
   id: string;
+  directory_id: string;
   raw_attributes: any;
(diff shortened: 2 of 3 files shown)


maxdeviant:
> Note: I haven't looked into whether `directory_id` should be present on all uses of the `User` interface (or if this "merged" type is only used in webhooks

The `directory_id` is a core property of the `User` type, so we can add it there 🙂 

It looks like it is missing on `Group` as well, if you'd like to add it there as well.

andyrichardson:
@maxdeviant thanks for the response!

Sorry to be second guessing you here, I'm looking at some of the example docs and the `User` and `Group` types don't show the `directory_id` on those entities.

Screenshot from 2021-11-09 10-14-17

Is it safe to assume that the examples are incorrectly missing the `directory_id` attribute on `User` and `Group` entities?

maxdeviant:
> @maxdeviant thanks for the response!
> 
> Sorry to be second guessing you here, I'm looking at some of the example docs and the `User` and `Group` types don't show the `directory_id` on those entities.
> 
> Screenshot from 2021-11-09 10-14-17
> 
> Is it safe to assume that the examples are incorrectly missing the `directory_id` attribute on `User` and `Group` entities?

Yep, those fields should be present in the Docs. We can get those added there!

maxdeviant (approved):

PR #517, opened 2021-11-08
Stringify payload in webhooks validation to resolve error

awolfden:
Add JSON.stringify to payload variable in computeSignature webhook method to resolve the 'Signature hash does not match the expected signature hash for payload' error


logangingerich (approved):
Thanks for catching!

PR #530, opened 2021-12-03
Update default tolerance to 3 minutes

MagsMagnoli:
According to the documentation, tolerances are between '3-5 minutes'. This library is instead validating against 180 milliseconds
This PR updates the default values to 3 minutes

### DIFF
--- a/src/webhooks/webhooks.ts
+++ b/src/webhooks/webhooks.ts
@@ -7,7 +7,7 @@ export class Webhooks {
     payload,
     sigHeader,
     secret,
-    tolerance = 180,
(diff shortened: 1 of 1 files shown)


willmanduffy:
Thanks for this @MagsMagnoli and great catch. We'll have this out in the next release.

willmanduffy (approved):

maxdeviant:
@MagsMagnoli This change is available in [v2.3.0](https://www.npmjs.com/package/@workos-inc/node/v/2.3.0) of the SDK.

PR #562, opened 2022-01-24
Export organization type definitions at top level

cacheonly:
I'm building a wrapper around the WorkOS client in TypeScript and noticed that types such as `Organization` and `CreateOrganizationOptions` are not exported by `@workos-inc/node`. Instead I had to import them from `@workos-inc/node/lib/organizations/interfaces`.


maxdeviant:
Thanks for the PR!

maxdeviant (approved):

PR #573, opened 2022-02-28
Add Organization ID parameter to list directories

sheldonvaughn:
Adds organization id parameter to the list directories pagination options


src/directory-sync/interfaces/list-directories-options.interface.ts
```diff
@@ -2,5 +2,6 @@ import { PaginationOptions } from '../../common/interfaces/pagination-options.in
 
 export interface ListDirectoriesOptions extends PaginationOptions {
   domain?: string;
+  organization?: string;
```
maxdeviant, line 5: This should be `organization_id` to match what the API is expecting:

```suggestion
  organization_id?: string;
```

PR #576, opened 2022-03-08
Add organization_id parameter to listDirectories method

sheldonvaughn:
resolves SUP-267 https://linear.app/workos/issue/SUP-267/node-sdk-add-organization-id-filter-to-list-directories-method


maxdeviant (approved):

src/directory-sync/interfaces/list-directories-options.interface.ts
```diff
@@ -2,5 +2,6 @@ import { PaginationOptions } from '../../common/interfaces/pagination-options.in
 
 export interface ListDirectoriesOptions extends PaginationOptions {
   domain?: string;
+  organization_id: string;
```
maxdeviant, line 5: This should be optional:

```suggestion
  organization_id?: string;
```

PR #584, opened 2022-03-28
Add SemVer description to README

sheldonvaughn:


README.md
```diff
@@ -33,6 +33,10 @@ import WorkOS from '@workos-inc/node';
 const workos = new WorkOS('sk_1234');
 ```
 
+## SDK Versioning
+
+For our SDKs WorkOS follows a Semantic Versioning process where all releases will have a version X.Y.Z (like 1.0.0) pattern wherein Z would be a bug fix (I.e. 1.0.1), Y would be a minor release (1.1.0) and X would be a major release (2.0.0). We permit any breaking changes to only be released in major versions and strongly recommend reading changelogs before making any major version upgrades.
```
maxdeviant, line 38: It might be good to link to the [SemVer](https://semver.org/) website in here.

sheldonvaughn: Thank you Marshall!

PR #611, opened 2022-05-19
Update Directory Sync Webhook Events types

lukemorales:
RESOLVES SDK-425

### DIFF
--- a/src/webhooks/interfaces/webhook-directory.interface.ts
+++ b/src/webhooks/interfaces/webhook-directory.interface.ts
@@ -1,10 +1,25 @@
-import { Directory } from '../../directory-sync/interfaces';
+export type WebhookDirectoryState =
+  | 'active'
+  | 'validating'
+  | 'invalid_credentials'
+  | 'inactive'
+  | 'deleting';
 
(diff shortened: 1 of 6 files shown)


src/webhooks/interfaces/webhook-directory.interface.ts
```diff
@@ -1,10 +1,25 @@
...
+  object: 'organization_domain';
+  id: string;
+  domain: string;
+}
+
+export interface WebhookDirectory {
+  object: 'directory';
+  id: string;
+  external_key: string;
+  type: string;
```
maxdeviant, line 18: Should this be a `DirectoryType`?

```suggestion
  type: DirectoryType;
```

maxdeviant: Ah, nevermind, I see we don't currently have a `DirectoryType` in the SDK.

maxdeviant (approved):

src/webhooks/interfaces/webhook-directory.interface.ts
lukemorales: @maxdeviant pushed the change

PR #615, opened 2022-05-23
Fix handling of `UnprocessableEntityException` responses

robframpton:
API currently can return `UnprocessableEntityException` with `code` and `message` properties. The Node SDK currently does not work with such responses, rather it always expects an `errors` property that it maps to generate a unique message.

This expands the `UnprocessableEntityException` error in the SDK to handle both cases.

I'm also removing a response type in the MFA API that can never be returned due to how the HTTP client handles non 200 status codes.

Should also be mentioned that this will also fix other exceptions outside of the MFA API as well.

### DIFF
--- a/src/common/exceptions/unprocessable-entity.exception.ts
+++ b/src/common/exceptions/unprocessable-entity.exception.ts
@@ -7,20 +7,22 @@ export class UnprocessableEntityException extends Error {
   readonly name: string = 'UnprocessableEntityException';
   readonly message: string = 'Unprocessable entity';
   readonly code?: string;
+  readonly requestID: string;
 
-  constructor(
-    {
-      code,
-      errors,
-      message,
-    }: {
-      code?: string;
-      errors?: UnprocessableEntityError[];
-      message?: string;
-    },
-    readonly requestID: string,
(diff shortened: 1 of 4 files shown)


src/common/exceptions/unprocessable-entity.exception.ts
```diff
@@ -5,16 +5,39 @@ import { UnprocessableEntityError } from '../interfaces';
...
+    {
+      code,
+      errors,
+      message,
+    }: {
+      code?: string;
+      errors?: UnprocessableEntityError[];
+      message?: string;
+    },
+    readonly requestID: string,
```
dliu-workos, line 21: why not roll requestId into the constructor object as well?

robframpton: I don't think there is a way to use the `readonly` property + constructor argument shorthand that way unfortunately.

robframpton: Updated by adding it to the object and just manually assigning property in the constructor.

dliu-workos (approved):

PR #616, opened 2022-05-23
v2.6.1

robframpton:
## Fixed

- Propagate `code` and `message` error properties for 422 responses
- Fixed TS types for `verifyFactor` response


awolfden (approved):
👍🏽

PR #622, opened 2022-06-06
Add idp_id, created_at and updated_at fields to Directory Group interface

logangingerich:
Update the Directory Group interface to add currently missing fields being returned by the API. Specifically `idp_id`, `created_at` and `updated_at`.


adamzinder (approved):
Looks good to me 👍

PR #624, opened 2022-06-07
Update connection enum list and bump version to 2.7.0

logangingerich:


pliao28 (approved):
Just double checking the package-lock.json file is suppose to remove all those packages, otherwise lgtm

PR #636, opened 2022-07-11
Add `auditLogs` module

robframpton:
Adds `auditLogs` module with `createEvent` method for emitting audit log events.

### DIFF
--- a/src/audit-logs/audit-logs.ts
+++ b/src/audit-logs/audit-logs.ts
@@ -1,0 +1,16 @@
+import { WorkOS } from '../workos';
+import { AuditLogEventOptions } from './interfaces';
+
+export class AuditLogs {
+  constructor(private readonly workos: WorkOS) {}
+
+  async createEvent(
+    organizationId: string,
+    event: AuditLogEventOptions,
+  ): Promise<void> {
+    await this.workos.post('/audit_logs/events', {
+      event,
+      organization_id: organizationId,
+    });
+  }
+}
--- a/src/audit-logs/audit-logs.ts
+++ b/src/audit-logs/audit-logs.ts
@@ -8,7 +8,7 @@ export class AuditLogs {
   constructor(private readonly workos: WorkOS) {}
 
   async createEvent(
-    { organization }: { organization: string },
+    organization: string,
     event: CreateAuditLogEventOptions,
     options: CreateAuditLogEventRequestOptions = {},
   ): Promise<void> {
--- a/src/audit-logs/audit-logs.ts
+++ b/src/audit-logs/audit-logs.ts
@@ -1,16 +1,24 @@
 import { WorkOS } from '../workos';
-import { AuditLogEventOptions } from './interfaces';
+import {
+  CreateAuditLogEventOptions,
+  CreateAuditLogEventRequestOptions,
+} from './interfaces';
 
 export class AuditLogs {
   constructor(private readonly workos: WorkOS) {}
 
   async createEvent(
-    organizationId: string,
-    event: AuditLogEventOptions,
+    { organization }: { organization: string },
+    event: CreateAuditLogEventOptions,
+    options: CreateAuditLogEventRequestOptions = {},
   ): Promise<void> {
-    await this.workos.post('/audit_logs/events', {
-      event,
-      organization_id: organizationId,
-    });
+    await this.workos.post(
+      '/audit_logs/events',
+      {
+        event,
+        organization_id: organization,
+      },
+      options,
+    );
   }
 }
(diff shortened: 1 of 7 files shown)


src/workos.ts
```diff
@@ -124,6 +128,13 @@ export class WorkOS {
                 error,
                 errorDescription,
               );
+            } else if (code && errors) {
+              throw new BadRequestException({
```
robframpton, line 132: I would much rather this be under `case 400: {`, but it breaks a unit test for `OauthException` ^.

We might be able to change it in the future, but we need to do some digging to ensure we don't break anything.

dliu-workos: Perhaps add a comment describing why we have this as a block instead of a `case 400:`

src/common/exceptions/bad-request.exception.ts
```diff
@@ -0,0 +1,36 @@
+export class BadRequestException extends Error {
+  readonly status: number = 400;
+  readonly name: string = 'BadRequestException';
+  readonly message: string = 'Bad request';
```
dliu-workos, line 4:
```suggestion
  readonly message: string = 'Bad Request';
```

src/audit-logs/audit-logs.ts
```diff
@@ -0,0 +1,16 @@
+import { WorkOS } from '../workos';
+import { AuditLogEventOptions } from './interfaces';
+
+export class AuditLogs {
+  constructor(private readonly workos: WorkOS) {}
+
+  async createEvent(
+    organizationId: string,
```
dliu-workos, line 8: Seems like in our other calls we put `organization_id` as a param inside the options

src/audit-logs/interfaces/audit-log-event-options.interface.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export interface AuditLogTarget {
+  id: string;
+  name?: string;
+  type: string;
+  metadata?: Record<string, string | number | boolean>;
+}
+
+export interface AuditLogEventOptions {
```
dliu-workos, line 15:
```suggestion
export interface AuditLogCreateEventOptions {
```

src/audit-logs/audit-logs.ts
```diff
@@ -0,0 +1,16 @@
+import { WorkOS } from '../workos';
+import { AuditLogEventOptions } from './interfaces';
+
+export class AuditLogs {
+  constructor(private readonly workos: WorkOS) {}
+
+  async createEvent(
```
dliu-workos, line 7: It'll be important to allow customers to pass an `idempotencyKey` like we do in the old API to prevent accidental creation of duplicate events.

src/audit-logs/interfaces/audit-log-event-options.interface.ts
dliu-workos: Or if this is the shape of the object itself, perhaps it should just be `AuditLogEvent` and `AuditLogCreateEv evntOptions` would house just the options?

src/audit-logs/audit-logs.ts
```diff
@@ -0,0 +1,16 @@
...
+    organizationId: string,
+    event: AuditLogEventOptions,
+  ): Promise<void> {
+    await this.workos.post('/audit_logs/events', {
+      event,
```
dliu-workos, line 12: I think we'll need to stringify the Date in the event. How do we handle dates elsewhere? Do we just send the ISOString or a timestamp (unix time?) I prefer unix time, but we'll probably want to be consistent with other APIs.

robframpton: Axios does that actually.

robframpton:
@dliu-workos here is an example of the updated request signature:

```typescript
workos.auditLogs.createEvent(
  { organization: 'org_123' }, 
  event, 
  {
    idempotencyKey: 'the-idempotency-key',
  }
);
```

dliu-workos:
> @dliu-workos here is an example of the updated request signature:
> 
> ```ts
> workos.auditLogs.createEvent(
>   { organization: 'org_123' }, 
>   event, 
>   {
>     idempotencyKey: 'the-idempotency-key',
>   }
> );
> ```

Not sure if the object gives much of anything (passing in `null` vs an empty object). Initially I thought we could add the `organization` as part of the `event` but it's not really a property of the `event`. Now I see why we have `organization` outside of the `event`.

How about swapping it to the original interface and when we want to make organization optional, we can overload the function?

```ts
createEvent(organization: string, event: CreateEventOptions, requestOptions: RequestOptions = {}): Promise<void>
createEvent(event:CreateEventOptions, requestOptions: RequestOptions = {}) {
   ....
}
```

robframpton:
In the event of making organization ID optional, we'll be replacing with a different primitive. Either way though this method signature should be fine though.

dliu-workos (approved):

PR #637, opened 2022-07-12
v2.8.0

dliu-workos:
Bumping version to 2.8.0 for release:
- Add `auditLogs` modules (preview) https://github.com/workos/workos-node/pull/636
- Updated endpoints used for MFA operations https://github.com/workos/workos-node/pull/633


maxdeviant:
We also need to update this test every time we bump the version: https://github.com/workos/workos-node/blob/168532a7de37603d01003457f820bd285c6c7f1f/src/sso/__snapshots__/sso.spec.ts.snap#L24

maxdeviant:
We need to bump the version here as well: https://github.com/workos/workos-node/blob/168532a7de37603d01003457f820bd285c6c7f1f/src/workos.ts#L27

maxdeviant (approved):

PR #638, opened 2022-07-13
Fix misleading 404 messages

dliu-workos:
Fixes misleading 404 messages returned from the API and refactors error mapping.

Before:
```
The requested path '/audit_logs/events' could not be found.
    err: {
     "type": "NotFoundException",
     "message": "The requested path '/audit_logs/events' could not be found.",
     "stack":
         NotFoundException: The requested path '/audit_logs/events' could not be found.
            at WorkOS.<anonymous> (....)
      "status": 404,
      "name": "NotFoundException"
    }
```

After:
```
Audit Log Validator not found: 'user.login_succeeded.1'.
    err: {
      "type": "NotFoundException",
      "message": "Audit Log Validator not found: 'user.login_succeeded.1'.",
      "stack":
          NotFoundException: Audit Log Validator not found: 'user.login_succeeded.1'.
              at WorkOS.handleAxiosError (...)
      "status": 404,
      "name": "NotFoundException",
      "code": "entity_not_found"
    }
```

PR #639, opened 2022-07-13
v2.8.1

dliu-workos:
Bumping to version 2.8.1 for release:
- https://github.com/workos/workos-node/pull/638


maxdeviant (approved):

package-lock.json
```diff
@@ -1,8 +1,5110 @@
 {
```
dliu-workos, line 1: seems like we can just remove this file. we have both a yarn.lock and a package-lock.json file

PR #642, opened 2022-07-14
Bump to 2.9.1

willmanduffy:
Bumping to version 2.9.1 for release:

* https://github.com/workos/workos-node/pull/640
* https://github.com/workos/workos-node/pull/641

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "2.8.1",
+  "version": "2.9.1",
   "name": "@workos-inc/node",
   "author": "WorkOS",
   "description": "A Node wrapper for the WorkOS API",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "2.8.1",
+  "version": "2.9.1",
```
maxdeviant, line 2: If we're bumping the minor version we should reset the patch:

```suggestion
  "version": "2.9.0",
```

willmanduffy: Don't let me write code at 7am.

maxdeviant: Putting the parental controls on your work machine as we speak: "No VS Code before 7am"

maxdeviant (approved):

PR #645, opened 2022-07-15
Add `createExport` and `getExport` to `auditLogs` module

robframpton:


src/audit-logs/interfaces/audit-log-export.interface.ts
```diff
@@ -0,0 +1,8 @@
+export interface AuditLogExport {
+  object: 'audit_log_export';
+  id: string;
+  state: string;
```
dliu-workos, line 4: do we have more granular states? `'complete' | 'failed' | 'in_progress | ...`

dliu-workos (approved):

PR #647, opened 2022-07-20
Add `actions`, `actors`, and `targets` filter to `createExport` method

robframpton:


src/audit-logs/interfaces/audit-log-export-options.interface.ts
```diff
@@ -1,5 +1,8 @@
 export interface AuditLogExportOptions {
+  actions?: string[];
```
maxdeviant, line 2: Seems like it would make more sense to have these grouped together at the end since they're all optional filters rather than sorting the fields alphabetically.

maxdeviant (approved):

erikengervall (approved):

PR #653, opened 2022-08-15
Allow customizing some Axios instance properties

### LINKED ISSUE #627: Allow customization of Axios instance
## Problem
I'm using this SDK in a project running on Cloudflare Workers which is not a native Node runtime. Axios expects either `XMLHttpRequest` or Node's `http` module to be available which causes an error since Cloudflare Workers only supports the standard [Fetch API](https://developers.cloudflare.com/workers/runtime-apis/fetch).

## Proposal
Axios accepts custom adapters to provide the underlying http client implementation. The adapter is specified within [Request Config](https://axios-http.com/docs/req_config). I would like to customize the Axios client instance by passing in additional options to the WorkOS constructor. This could be done by passing in a Axios instance or the Request Config used to create the Axios instance. This would also be useful to create mock implementation for testing.

For example,
```javascript
new WorkOS(key, {
  axios: axios.create({ adapter: myFetchAdapter })
});
```

### LINKED ISSUE #652: Node versions before 10 do not support new WorkOS Root CA
Legacy Node versions, which many apps still run on, bundle a set of root CAs and do not support the new root from Let's Encrypt for WorkOS's https://api.workos.com, which makes the service incompatible with legacy software as of August 10, 2022.

The error message returned from Axios is `certificate has expired`.

The fix requires allowing the consumer to pass an `https.Agent` instance into the WorkOS constructor which includes the valid root CA.

At the current time, the WorkOS SSL certificate indicates the chain is `ISRG Root X2 > E1 > *.workos.com`. The root CA file can be obtained from from [Let's Encrypt](https://letsencrypt.org/certificates/) directly. This is the "Self-Signed" pem file (cross-signed does *not* work in our testing) and is named `isrg-root-x2.pem`.

A simple test to see if your Node version supports the new certificate:

```js
const axios = require('axios');

axios.get('https://api.workos.com')
  .then(function (response) {
    console.log(response.data);
  })
  .catch(function (error) {
    console.log(error.message);
  });
```

If this logs out `certificate has expired` then your Node version does not support the new SSL certificate.

jdforsythe:
Fixes #652
Fixes #627

This adds three specific Axios options to `WorkOSOptions`: `adapter`, `httpsAgent`, and `proxy`.

The first two options address the issues listed above. The `proxy` may be needed in some instances. I specifically used `Pick` to pull only these properties to keep the consumer from messing up the functionality by passing in a different `baseUrl` or other properties.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,32 @@ import WorkOS from '@workos-inc/node';
 const workos = new WorkOS('sk_1234');
 ```
 
+## Axios customization
+
+Some users may need to customize the underlying Axios instance. For instance, you may need to pass an `https.Agent` with a root CA to allow older Node versions to call the API, or you may be running on a non-Node runtime and need to pass in a custom `adapter`.
+
+You can pass in one of three Axios options in the `WorkOSOptions` object:
+
+- `adapter`
+- `httpsAgent`
+- `proxy`
+
+Example of using a custom root CA for legacy Node support:
+
+```ts
+import { Agent } from 'https';
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+import WorkOS from '@workos-inc/node';
+
+const rootCA = readFileSync(join(__dirname, './isrg-root-x2.pem'));
+
+const agent = new Agent({ ca: rootCA });
+
+const workos = new WorkOS('sk_1234', { axios: { httpsAgent: agent } });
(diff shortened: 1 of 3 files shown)


maxdeviant:
Thanks for the PR!

We've implemented this separately in #657.

PR #655, opened 2022-08-16
Add support for idempotency keys when creating organizations

dliu-workos:


src/organizations/interfaces/create-organization-options.interface.ts
```diff
@@ -1,5 +1,12 @@
+import { PostOptions } from '../../common/interfaces';
+
 export interface CreateOrganizationOptions {
   name: string;
   allow_profiles_outside_organization?: boolean;
   domains?: string[];
 }
+
+export type CreateOrganizationRequestOptions = Pick<
```
maxdeviant, line 9: We can use an `interface` type here:

```suggestion
export interface CreateOrganizationRequestOptions extends Pick<
```

maxdeviant (approved):

PR #656, opened 2022-08-18
Uncover private methods in webhooks class

edignot:
In addition to the previous public method in the Webhooks class, constructEvent, we also uncover the private methods in this class:

verifyHeader
getTimestampAndSignatureHash
computeSignature
secureCompare

Followed Ruby PR as an example: https://github.com/workos/workos-ruby/pull/171


logangingerich (approved):
Looking perfect, great work!

PR #657, opened 2022-08-18
Allow customizing Axios client

maxdeviant:
This PR adds support for customizing the underlying Axios client.

Resolves SDK-577.

### DIFF
--- a/src/common/interfaces/workos-options.interface.ts
+++ b/src/common/interfaces/workos-options.interface.ts
@@ -1,5 +1,8 @@
+import { AxiosRequestConfig } from 'axios';
+
 export interface WorkOSOptions {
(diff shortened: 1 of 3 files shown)


mthadley (approved):

herbertpimentel:
Do you have any instructions about how to customize is made, what should we do in order to make the error go way 

{
"code": 500,
"message": "certificate has expired"
}

maxdeviant:
> Do you have any instructions about how to customize is made, what should we do in order to make the error go way
> 
> { "code": 500, "message": "certificate has expired" }

There are some more details here: https://github.com/workos/workos-node/issues/652#issuecomment-1215131043

PR #670, opened 2022-09-25
Add method for directory user primary email

edignot:
Adds method on the directory sync class to get the primary email out of the email array.

### DIFF
--- a/src/directory-sync/directory-sync.ts
+++ b/src/directory-sync/directory-sync.ts
@@ -59,8 +59,8 @@ export class DirectorySync {
     return data;
   }
 
-  getPrimaryEmail(user: UserWithGroups ): string | undefined {
-    const primaryEmail = (user.emails || []).find(email => email.primary);
+  getPrimaryEmail(user: UserWithGroups): string | undefined {
+    const primaryEmail = (user.emails || []).find((email) => email.primary);
     return primaryEmail && primaryEmail.value;
   }
 }
--- a/src/directory-sync/directory-sync.ts
+++ b/src/directory-sync/directory-sync.ts
@@ -58,4 +58,9 @@ export class DirectorySync {
     const { data } = await this.workos.get(`/directory_groups/${group}`);
     return data;
   }
+
+  getPrimaryEmail(user: UserWithGroups ): string | undefined {
+    const primaryEmail = (user.emails || []).find(email => email.primary);
+    return primaryEmail && primaryEmail.value;
+  }
 }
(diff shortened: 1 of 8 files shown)


src/directory-sync/directory-sync.ts
```diff
@@ -58,4 +58,9 @@ export class DirectorySync {
     const { data } = await this.workos.get(`/directory_groups/${group}`);
     return data;
   }
+
+  getPrimaryEmail(user: UserWithGroups): string | undefined {
```
maxdeviant, line 62: Could we make this a free function instead?

There's no reason why it needs to be tied to the class instance.

edignot: @maxdeviant, would you mind clarifying where exactly (which file) this function needs to be added then? workos.js is also a class, so then it will get tied to that class as I understand.

maxdeviant: We can export it as a function at the top of this file:

```ts
export function getPrimaryEmail(user: User): string | undefined {
  // ...
}

export class DirectorySync {
  // ...
}
```

maxdeviant: We'll also need to make sure we update the root barrel file to ensure that it is exposed outside of the package: https://github.com/workos/workos-node/blob/8a118e06226e7075fed30261a568b9363fddd71a/src/index.ts#L7

src/directory-sync/utils/get-primary-email.ts
```diff
@@ -0,0 +1,6 @@
+import { UserWithGroups } from '../interfaces/user.interface';
+
+export function getPrimaryEmail(user: UserWithGroups): string | undefined {
```
maxdeviant, line 3: We can make this take just a plain `User`.

`UserWithGroups` is a supertype of `User`, so it can also be passed since it structurally satisfies the `User` type.

src/directory-sync/utils/get-primary-email.ts
```diff
@@ -0,0 +1,6 @@
...
+  const primaryEmail = (user.emails || []).find((email) => email.primary);
+  return primaryEmail && primaryEmail.value;
```
maxdeviant, line 5: We should be able to use optional chaining here instead:

```suggestion
  return primaryEmail?.value;
```

src/directory-sync/utils/get-primary-email.ts
```diff
@@ -0,0 +1,6 @@
+import { User } from '../interfaces/user.interface';
+
+export function getPrimaryEmail(user: User): string | undefined {
+  const primaryEmail = (user.emails || []).find((email) => email.primary);
```
maxdeviant, line 4: We should be able to use optional chaining here instead of adding a default value:

```suggestion
  const primaryEmail = user.emails?.find((email) => email.primary);
```

maxdeviant (approved):

PR #675, opened 2022-10-11
Add job title to directory user

blairlunceford:
Adds job title to directory user


src/directory-sync/directory-sync.spec.ts
```diff
@@ -56,6 +56,7 @@ describe('DirectorySync', () => {
     groups: [groupResponse],
     idp_id: 'idp_foo',
     last_name: 'Snow',
+    job_title: 'Knight of the Watch',
```
ams11, line 59: 😂  ⛄⚔️

src/directory-sync/interfaces/user.interface.ts
```diff
@@ -20,6 +20,7 @@ export interface User<
   }[];
   username: string;
   last_name: string;
+  job_title: string;
```
ams11, line 23: Do we not need any "setters" for these fields in Node - is it smart enough to just extract them from the payload auto-magically? I'm assuming so since the tests pass 😅, but just wanted to confirm (since I thought I'd find something like https://github.com/workos/workos-ruby/pull/178/files#diff-df42219ae143db61d39dd9502412fe14ef802a013750d3f599395b19540e3056R27 in here too)

ams11 (approved):
Approving, since I think/assume this all works now, but just had a question clarifying that we don't need any sort of a setter method...

src/directory-sync/interfaces/user.interface.ts
maxdeviant: The underlying objects used by the SDK are just Plain Old JavaScript Objects (POJOs), so the typings here are just applied on top of the object (which may not actually match what is there at runtime).

So no, we don't need to specify any "setters".

maxdeviant, line 23: This should be marked as nullable:

```suggestion
  job_title: string | null;
```

blairlunceford: 👍 thanks! Made this change.

maxdeviant (approved):

PR #684, opened 2022-11-17
Create a PR template

vladmoroz:
Setting up a PR template reminding to update the docs so that:
- We don't forget to update the docs when developer API changes
- If needed, the DX team can provide guidance on creating a coherent API across SDKs


danerwilliams (approved):
lgtm!

PR #704, opened 2023-02-16
Add groups attribute to user profile

blairlunceford:
## Description

Adds groups attribute to user profile. Available behind a feature flag.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -16,6 +16,7 @@ global_job_config:
   prologue:
     commands:
       - checkout
+      - nvm install
       - nvm use
       - bash ~/workos-node/bin/restore-or-install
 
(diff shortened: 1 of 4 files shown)


blairlunceford:
Had to update CI pipeline - this happened in the mono repo here: https://work-os.slack.com/archives/CG875BV3J/p1675684580898109

ameesha (commented):
looks fine but do we want to put these in the SDKs and accessible already while the flag is off?

blairlunceford:
> looks fine but do we want to put these in the SDKs and accessible already while the flag is off?

This is necessary to allow a beta test - a customer is using this SDK, so we need to support it here for them to get the groups attribute. It should be backwards compatible - I added a test for when groups aren't passed, and we don't see it in the profile.

ameesha (approved):

PR #721, opened 2023-04-06
Remove `query-string` dependency

mthadley:
## Description

Removes the dependency on `query-string` and replaces it with built-in Node API's.

[We've received automated version bumps for `query-string`](https://github.com/workos/workos-node/pull/696), but given the new version requires ESM (which we don't currently use), and how small our usage of `query-string` is, it seems easier to remove it.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/sso/sso.spec.ts
+++ b/src/sso/sso.spec.ts
@@ -126,7 +126,7 @@ describe('SSO', () => {
           });
 
           expect(url).toMatchInlineSnapshot(
-            `"https://api.workos.com/sso/authorize?client_id=proj_123&connection=connection_123&domain_hint=lyft.com&redirect_uri=example.com%2Fsso%2Fworkos%2Fcallback&response_type=code&state=custom%20state"`,
+            `"https://api.workos.com/sso/authorize?client_id=proj_123&connection=connection_123&domain_hint=lyft.com&redirect_uri=example.com%2Fsso%2Fworkos%2Fcallback&response_type=code&state=custom+state"`,
           );
         });
       });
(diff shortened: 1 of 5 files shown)


src/sso/sso.spec.ts
```diff
@@ -144,7 +144,7 @@ describe('SSO', () => {
           });
 
           expect(url).toMatchInlineSnapshot(
-            `"https://api.workos.com/sso/authorize?client_id=proj_123&connection=connection_123&login_hint=foo%40workos.com&redirect_uri=example.com%2Fsso%2Fworkos%2Fcallback&response_type=code&state=custom%20state"`,
+            `"https://api.workos.com/sso/authorize?client_id=proj_123&connection=connection_123&login_hint=foo%40workos.com&redirect_uri=example.com%2Fsso%2Fworkos%2Fcallback&response_type=code&state=custom+state"`,
```
mthadley, line 147: `query-string` previously encoded spaces as `%20`, while `URLSearchParams` uses a `+`.

The WorkOS supports both, [though it seems like `+` is the preferred method](https://stackoverflow.com/questions/1211229/in-a-url-should-spaces-be-encoded-using-20-or).

src/sso/sso.ts
```diff
@@ -9,6 +8,21 @@ import { ListConnectionsOptions } from './interfaces/list-connections-options.in
 import { ProfileAndToken } from './interfaces/profile-and-token.interface';
 import { Profile } from './interfaces/profile.interface';
 
+const toQueryString = (options: Record<string, string | undefined>): string => {
+  const searchParams = new URLSearchParams();
+  const keys = Object.keys(options).sort();
```
mthadley, line 13: `query-string` was also sorting the params under the hood. This at least keeps the output deterministic, which is a good thing for the existing test suite.

maxdeviant (approved):
Beautiful!

PR #741, opened 2023-06-01
Add Events API beta

alisherry:
## Description
Adding beta support for the Events API.

## Documentation

```
[ X ] Yes
```

### DIFF
--- a/src/events/events.ts
+++ b/src/events/events.ts
@@ -1,0 +1,17 @@
+import { WorkOS } from '../workos';
+import { Event } from './interfaces/event.interface';
+import { List } from '../common/interfaces/list.interface';
+import {GetEventOptions} from './interfaces';
+
+export class Events{
+  constructor(private readonly workos: WorkOS) {}
+
+  async getEvents(
+    options: GetEventOptions 
+  ): Promise<List<Event>>{
+    const { data } = await this.workos.get(`/events`, {
+      query: options,
+    });
+    return data;
+  }
+}
--- a/src/events/events.ts
+++ b/src/events/events.ts
@@ -1,12 +1,12 @@
 import { WorkOS } from '../workos';
 import { Event } from './interfaces/event.interface';
 import { List } from '../common/interfaces/list.interface';
-import { GetEventOptions } from './interfaces';
+import { ListEventOptions } from './interfaces';
 
 export class Events {
   constructor(private readonly workos: WorkOS) {}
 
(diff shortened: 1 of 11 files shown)


src/events/events.ts
```diff
@@ -0,0 +1,15 @@
+import { WorkOS } from '../workos';
+import { Event } from './interfaces/event.interface';
+import { List } from '../common/interfaces/list.interface';
+import { GetEventOptions } from './interfaces';
+
+export class Events {
+  constructor(private readonly workos: WorkOS) {}
+
+  async getEvents(options: GetEventOptions): Promise<List<Event>> {
```
tribble, line 9: To match our other APIs, I think this should be `listEvents`

src/events/interfaces/event.interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface Event {
+  id: string;
+  event: string;
+  created_at: string;
+  data: Record<string, unknown>;
+}
```
tribble, line 6: I think we can be more specific here, like we are [with webhook types](https://github.com/workos/workos-node/blob/main/src/webhooks/interfaces/webhook.interface.ts).

src/events/interfaces/get-events-options.interface.ts
```diff
@@ -0,0 +1,7 @@
+export interface GetEventOptions {
+  events?: string[];
```
tribble, line 2: Can we be more specific here? There are a finite number of valid event types.

tribble (commented):
Technically Events is a list endpoint and I think we should name our SDK method to match. I also think we can be more strict with the types.

ospahiu:
Do we usually do tests for the node SDK?

alisherry:
@ospahiu I don't see any outside of more generic tests for particular products, might make sense to add some more tests with options but I don't think we have any other E2E ones

src/events/interfaces/event.interface.ts
```diff
@@ -0,0 +1,106 @@
...
+  | 'connection.deactivated'
+  | 'connection.deleted'
+  | 'dsync.activated'
+  | 'dsync.deactivated'
+  | 'dsync.deleted'
+  | 'dsync.group.created'
+  | 'dsync.group.deleted'
+  | 'dsync.group.updated'
+  | 'dsync.group.user_added'
+  | 'dsync.group.user_removed'
```
tribble, line 103: 💯

tribble (approved):

PR #744, opened 2023-06-09
Add `Events` API  beta to v2.x

alisherry:
## Description
Adding `Events` API beta 
## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/audit-trail/interfaces/event.interface.ts
+++ b/src/audit-trail/interfaces/event.interface.ts
@@ -1,10 +1,10 @@
-export interface EventAction {
+export interface AuditTrailEventAction {
   object: 'event_action';
   id: string;
   name: string;
 }
 
-export interface Event {
+export interface AuditTrailEvent {
   object: 'event';
   id: string;
   group: string;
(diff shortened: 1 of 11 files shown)


src/audit-trail/interfaces/event.interface.ts
```diff
@@ -1,10 +1,10 @@
-export interface EventAction {
+export interface AuditTrailEventAction {
   object: 'event_action';
   id: string;
   name: string;
 }
 
-export interface Event {
```
tribble, line 7: Do we export these types? If someone were to update the SDK, and they reference `Event` from  their code, would this make their TS build fail? Or start picking up the new `Event` type?

tribble (commented):
Since this a minor version, and really, I should have caught this in the previous review, could this impact user's TS builds with the rename of `Event` to `AuditTrailEvent` and a new object shape taking over `Event`?

src/audit-trail/interfaces/event.interface.ts
alisherry: Hmm, yeah this would be a breaking change. Perhaps it makes sense to wait to update Audit Trails for the next major update.

PR #757, opened 2023-07-05
Devbox

AshGodfrey:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/devbox.lock
+++ b/devbox.lock
@@ -1,6 +1,11 @@
 {
   "lockfile_version": "1",
   "packages": {
+    "nodePackages.eslint@8.41.0": {
+      "last_modified": "2023-05-25T03:54:59Z",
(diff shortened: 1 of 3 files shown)


devbox.lock
```diff
@@ -0,0 +1,20 @@
+{
+  "lockfile_version": "1",
+  "packages": {
+    "nodePackages.eslint@8.41.0": {
```
maxdeviant, line 4: Is the lockfile up-to-date with `devbox.json`?

It looks like we have `nodePackages.eslint` referenced here, but it's not in the `packages` array (and probably doesn't need to be, based on our Slack discussion).

AshGodfrey: fixed - also went ahead and added the linter for SDK consistency.

maxdeviant (approved):

PR #762, opened 2023-07-18
Add UserManagement APIs

cmatheson:
## Description

Adding the new UserManagement API endpoints to the SDK.

## Documentation

```
[X] Yes
```


src/user-management/interfaces/user.interface.ts
```diff
@@ -0,0 +1,27 @@
+// TODO
+// - should User be defined as `ManagedUser | UnmanagedUser` ?
```
cmatheson, line 2: thoughts on this? typescript can model this nicer than having a bunch of optional fields, but leaving the User interface like this is probably easier for typescript novices?

maxdeviant: I think we should use a union type here.

src/users/fixtures/list-users.json
```diff
@@ -0,0 +1,25 @@
...
+    }
+  ],
+  "list_metadata": {
+    "before": null,
+    "after": null
+  },
+  "listMetadata": {
+    "before": null,
+    "after": null
+  }
```
oliverzheng, line 24: I believe this is no longer a thing after workos/workos#21671.

src/users/interfaces/challenge-response.interface.ts
```diff
@@ -0,0 +1,6 @@
+import { User } from './user.interface';
+
+export interface ChallengeResponse {
```
oliverzheng, line 3: Would it be clearer to have

```suggestion
export interface EmailVerificationChallengeResponse {
```

src/users/interfaces/user.interface.ts
```diff
@@ -0,0 +1,34 @@
...
+  identities: Identity[];
+}
+
+interface BaseUser {
+  object: 'user';
+  id: string;
+  email: string;
+  first_name?: string;
+  last_name?: string;
+  email_verified_at?: string;
```
oliverzheng, line 20: This is only a property on the `UnmanagedUser`.

src/users/interfaces/user.interface.ts
```diff
@@ -0,0 +1,34 @@
+export type User = ManagedUser | UnmanangedUser;
+
+interface ManagedUser extends BaseUser {
+  user_type: 'managed';
+  identities: [Identity];
+  sso_profile_id: string;
+}
+
+interface UnmanangedUser extends BaseUser {
```
oliverzheng, line 9: There's now `google_oauth_profile_id` on the unmanaged user as well.

src/users/interfaces/add-user-to-organization-options.interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface AddUserToOrganizationOptions {
+  id: string;
```
oliverzheng, line 2: Perhaps this can make it clearer the object for this ID?

```suggestion
  user_id: string;
```

src/users/interfaces/remove-user-from-organization-options.interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface RemoveUserFromOrganizationOptions {
+  id: string;
```
oliverzheng, line 2: Same here, since there are two ids.

```suggestion
  user_id: string;
```

src/users/fixtures/list-users.json
mthadley: Yep, we'll need to make sure the SDK doesn't expect the legacy pagination metadata field, since it won't be sent for these new APIs.

src/users/interfaces/challenge-response.interface.ts
cmatheson: I'm using the ChallengeResponse for the password challenge response too (I could separate those but they are identical for now?).

src/users/interfaces/user.interface.ts
cmatheson: Ohh, still an optional field on the UnmanagedUser i assume?

src/users/interfaces/add-user-to-organization-options.interface.ts
cmatheson: ok i started out with user_id and switched to `id` to match the api docs, but i agree it's more clear like this, will fix.

cmatheson: (maybe we should update the docs too)

src/users/interfaces/user.interface.ts
oliverzheng: Yep!

src/users/interfaces/challenge-response.interface.ts
oliverzheng: Yeah maybe separate types to denote that the response data are only incidentally and not purposely the same.

src/users/interfaces/add-user-to-organization-options.interface.ts
oliverzheng: Hmm the docs have the context that `:id` is part of the URL which is like 10 pixels above the definition of `id`. Having it be `:user_id` just for this endpoint would make it differ from other user endpoints that are just `:id`, and from what I can tell, all our point-load APIs have just `:id` in the URL for the docs.

I think it's ok in this case if the SDK deviates from the API docs, since there's no context within the SDK which ID is used where at the function signature level.

src/users/fixtures/list-users.json
cmatheson: thanks, my branch was out of date. fixed.

src/users/interfaces/challenge-response.interface.ts
cmatheson: separated!

src/users/interfaces/add-user-to-organization-options.interface.ts
cmatheson: ok, i've replaced all the `id`s with `user_id` for consistency.

cmatheson:
(converted this to a draft since we don't want to merge this into `main` yet).

oliverzheng (approved):

src/users/users.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  async listUsers(options?: ListUsersOptions): Promise<List<User>> {
+    const { data } = await this.workos.get('/users', {
+      query: options,
+    });
+    return data;
+  }
+
+  async createUnmanagedUser(
```
oliverzheng, line 37: Perhaps this can be called `createUser` since that's what the API endpoint really is.

src/users/users.ts
```diff
@@ -0,0 +1,135 @@
...
+    payload: CreateUnmanagedUserOptions,
+  ): Promise<User> {
+    const { data } = await this.workos.post('/users', payload);
+
+    return data;
+  }
+
+  async authenticateUnmanagedUser(
```
oliverzheng, line 45: Perhaps be more specific with the name:

```suggestion
  async authenticateUserWithPassword(
```

Because OAuth users are unmanaged but don't go through this endpoint.

src/users/users.ts
```diff
@@ -0,0 +1,135 @@
...
+    payload: AuthenticateUnmanagedUserOptions,
+  ): Promise<AuthenticationResponse> {
+    const { data } = await this.workos.post('/users/authentications', payload);
+    return data;
+  }
+
+  async authenticateManagedUser(
```
oliverzheng, line 52: Similar to the password one, the authentication mechanism is actually what separate this endpoint apart, not the type of user it returns.

```suggestion
  async authenticateUserWithToken(
```

src/users/interfaces/authenticate-managed-user-options.interface.ts
```diff
@@ -0,0 +1,8 @@
+export interface AuthenticateManagedUserOptions {
+  client_id: string;
+  client_secret: string;
+  grant_type: 'authorization_code';
```
oliverzheng, line 4: I think these two do not need to be specified here. One is a constant and the other is already specified in the SDK client instantiation. See how the SSO token endpoint fills these in:

https://github.com/workos/workos-node/blob/ddcd0ac0330e637a9b29b6bbfbf0c8e06b95e33b/src/sso/sso.ts#L83-L84

cmatheson: ah, great point :+1:

cmatheson: done.

src/users/users.ts
cmatheson: done

src/users/users.ts
cmatheson: done.

src/users/users.ts
cmatheson: done

PR #763, opened 2023-07-19
Update `main` from `v2.x` branch

maxdeviant:
## Description

This PR merges the changes from the `v2.x` branch into `main`.

## Documentation

```
[ ] Yes
```


package-lock.json
maxdeviant, line 1: We use Yarn for the SDK, so we don't have a `package-lock.json` file.

src/events/events.spec.ts
```diff
@@ -1,7 +1,5 @@
 import axios from 'axios';
-
 import MockAdapater from 'axios-mock-adapter';
```
maxdeviant, line 2: Noticed we have a typo here.

It seems to be more widespread, so will fix it in a follow-up PR.

cmatheson (approved):

PR #767, opened 2023-07-20
Merge main into v3.0.0

cmatheson:
## Description

Just getting the v3.0.0 branch current before tackling the items listed in https://www.notion.so/workos/Features-for-the-Next-Major-SDK-Release-4a425188c6b94d46a728e42abc509a62#c2e47b7921e14cc4adf649158c287f56.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,19 +34,19 @@
     "prepublishOnly": "yarn run build"
   },
   "dependencies": {
-    "axios": "1.3.6",
+    "axios": "1.4.0",
     "pluralize": "8.0.0"
   },
   "devDependencies": {
-    "@types/jest": "29.5.1",
-    "@types/node": "14.18.43",
-    "@types/pluralize": "0.0.29",
-    "axios-mock-adapter": "1.21.4",
-    "jest": "29.5.0",
+    "@types/jest": "29.5.3",
+    "@types/node": "14.18.53",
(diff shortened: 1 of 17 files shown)


maxdeviant:
> Just getting the v3.0.0 branch current

Is the "v3.0.0 branch" `main` in this case?

The PR title is a little confusing, since we're merging this PR into `main`.

cmatheson:
> The PR title is a little confusing, since we're merging this PR into main.

Ohh, it should have been `v3.0.0` (I was hoping `gh` would use my tracking branch as the branch to merge into, but it needed an edit).

maxdeviant:
> > The PR title is a little confusing, since we're merging this PR into main.
> 
> Ohh, it should have been `v3.0.0` (I was hoping `gh` would use my tracking branch as the branch to merge into, but it needed an edit).

Seems like we should just use `main` for the next major version release (v3) instead of creating a new branch to track it.

And then we just keep the `v2.x` around if we need to release new `v2.x.x` versions in the meantime.

cmatheson:
> Seems like we should just use main for the next major version release (v3) instead of creating a new branch to track it.

> And then we just keep the v2.x around if we need to release new v2.x.x versions in the meantime.

I like that idea better :+1:.

PR #769, opened 2023-07-20
Standardize API naming

cmatheson:
## Description

- rename `accessToken` to `access_token`
- Make SSO interface properties snake_case
- SSO.getAuthorizationURL -> SSO.getAuthorizationUrl
- Make MFA interfaces use snake_case
- Make admin portal interfaces use snake_case
- Make passwordless interfaces use snake_case
- Rename User to DirectoryUser

## Documentation

```
[X] Yes
```

### DIFF
--- a/src/directory-sync/interfaces/directory_user.interface.ts
+++ b/src/directory-sync/interfaces/directory_user.interface.ts
@@ -2,7 +2,7 @@ import { Group } from './group.interface';
 
 export type DefaultCustomAttributes = Record<string, unknown>;
 
-export interface User<
+export interface DirectoryUser<
   TCustomAttributes extends object = DefaultCustomAttributes,
   TRawAttributes = any,
 > {
@@ -24,8 +24,8 @@ export interface User<
   state: 'active' | 'inactive' | 'suspended';
 }
 
-export interface UserWithGroups<
+export interface DirectoryUserWithGroups<
   TCustomAttributes extends object = DefaultCustomAttributes,
-> extends User<TCustomAttributes> {
+> extends DirectoryUser<TCustomAttributes> {
   groups: Group[];
 }
--- a/src/mfa/interfaces/challenge-factor-options.ts
+++ b/src/mfa/interfaces/challenge-factor-options.ts
@@ -1,8 +1,4 @@
-export type ChallengeFactorOptions =
-  | {
-      authenticationFactorId: string;
-    }
-  | {
-      authenticationFactorId: string;
-      smsTemplate: string;
-    };
+export interface ChallengeFactorOptions {
+  authentication_factor_id: string;
+  sms_template?: string;
+}
--- a/src/common/interfaces/get-options.interface.ts
+++ b/src/common/interfaces/get-options.interface.ts
@@ -1,4 +1,4 @@
 export interface GetOptions {
   query?: Record<string, any>;
-  accessToken?: string;
+  access_token?: string;
 }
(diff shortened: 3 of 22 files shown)


src/directory-sync/interfaces/directory_user.interface.ts
```diff
@@ -24,8 +24,8 @@ export interface User<
   state: 'active' | 'inactive' | 'suspended';
 }
 
-export interface UserWithGroups<
+export interface DirectoryUserWithGroups<
```
cmatheson, line 27: is there any reason to keep this separate `DirectoryUserWithGroups` interface around? seems like all uses of `DirectoryUser` are actually `DirectoryUserWithGroups`

maxdeviant:
If we're standardizing on the casing, should we be standardizing on camelCase and then converting to and from snake_case at the API boundary?

It's a bit odd that we have any snake_case identifiers in the SDK's public API, as it's not a standard casing for Node packages.

src/mfa/interfaces/challenge-factor-options.ts
```diff
@@ -1,8 +1,4 @@
-export type ChallengeFactorOptions =
-  | {
-      authenticationFactorId: string;
-    }
-  | {
-      authenticationFactorId: string;
-      smsTemplate: string;
-    };
+export interface ChallengeFactorOptions {
```
cmatheson, line 1: pretty sure this is equivalent but feel free to disagree!

maxdeviant: I think we should keep it as a union type, as these are mutually exclusive options that could diverge more longer-term.

cmatheson:
> If we're standardizing on the casing, should we be standardizing on camelCase and then converting to and from snake_case at the API boundary?

Hmm, the upgrade path for existing people is nicer if we continue to use snake_case, but maybe that's not a good enough reason to keep preserve snake case.

maxdeviant:
> > If we're standardizing on the casing, should we be standardizing on camelCase and then converting to and from snake_case at the API boundary?
> 
> Hmm, the upgrade path for existing people is nicer if we continue to use snake_case, but maybe that's not a good enough reason to keep preserve snake case.

A couple of thoughts here from the lens of us wanting to eventually end up using camelCase across the board:

- There are some places where we're currently using camelCase that are being rewritten to snake_case in this PR. This seems like a step backwards for these names
- Newer APIs (User Management) should be introduced using camelCase so that it doesn't involve a migration later
- Short of writing and shipping a codemod, there will have to be some point where consumers switch over to the camelCased APIs, so I don't see a strong reason why that switch shouldn't take place in v3

oliverzheng:
I also prefer camelCasing in Node SDKs. What do we do with API responses that are snake_cased? Without doing anything about that, we'd have inconsistency within the usage of the SDK itself, where callsites would pass in camelCasedParams and get back snake_cased_properties all within the same code-block. 

As a point of reference, Stripe's Node SDK uses snake_casing for both params and responses: https://stripe.com/docs/api/credit_notes/create?lang=node

maxdeviant:
> I also prefer camelCasing in Node SDKs. What do we do with API responses that are snake_cased? Without doing anything about that, we'd have inconsistency within the usage of the SDK itself, where callsites would pass in camelCasedParams and get back snake_cased_properties all within the same code-block.

We could add a translation layer where we translate the snake_cased responses to camelCase. This is what our other SDKs do as well (they just typically have built-in support for it in their JSON serialization library).

We'll already need to do this anyways for normalizing some of the enum values (mentioned in the Notion doc about SDK changes).

cmatheson:
closing this branch (we should do camelCase instead of snake_case as done here).

PR #783, opened 2023-08-03
Add default order

AshGodfrey:
## Description

Passes in 'desc' as the default order when none is provided. 

## Documentation

```
[ ] Yes
```


src/webhooks/fixtures/webhook.json
```diff
@@ -1 +1,54 @@
...
+{
```
willmanduffy, line 1: What's the change in here?

src/portal/fixtures/generate-link-invalid.json
```diff
@@ -1 +1 @@
-{ "message": "Could not find an organization with the id, bogus-id."}
+{ "message": "Could not find an organization with the id, bogus-id." }
```
willmanduffy, line 1: This change makes sense, but curious why it's being included in this diff?

PR #788, opened 2023-08-04
Add auto-pagination

AshGodfrey:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/common/interfaces/index.ts
+++ b/src/common/interfaces/index.ts
@@ -7,3 +7,4 @@ export * from './put-options.interface';
 export * from './unprocessable-entity-error.interface';
 export * from './workos-options.interface';
 export * from './workos-response-error.interface';
+export * from './pagination-options.interface';
--- a/src/common/utils/fetch-and-deserialize.ts
+++ b/src/common/utils/fetch-and-deserialize.ts
@@ -1,0 +1,25 @@
+import { WorkOS } from '../../workos';
+import { List, ListResponse, PaginationOptions } from '../interfaces';
+import { deserializeList } from '../serializers';
+
+function setDefaultOptions(options?: PaginationOptions): PaginationOptions {
+  return {
+    ...options,
+    order: options?.order || 'desc',
+  };
+}
+
+export async function fetchAndDeserialize<T, U>(
+  workos: WorkOS,
+  endpoint: string,
+  deserializeFn: (data: T) => U,
+  options?: PaginationOptions,
+): Promise<List<U>> {
+  const finalOptions = setDefaultOptions(options);
+
+  const { data } = await workos.get<ListResponse<T>>(endpoint, {
+    query: finalOptions,
+  });
+
+  return deserializeList(data, deserializeFn);
+}
--- a/src/common/utils/fetch-and-deserialize.ts
+++ b/src/common/utils/fetch-and-deserialize.ts
@@ -2,24 +2,22 @@ import { WorkOS } from '../../workos';
 import { List, ListResponse, PaginationOptions } from '../interfaces';
 import { deserializeList } from '../serializers';
 
-function setDefaultOptions(options?: PaginationOptions): PaginationOptions {
+const setDefaultOptions = (options?: PaginationOptions): PaginationOptions => {
   return {
     ...options,
     order: options?.order || 'desc',
   };
-}
+};
 
-export async function fetchAndDeserialize<T, U>(
+export const fetchAndDeserialize = async <T, U>(
   workos: WorkOS,
   endpoint: string,
   deserializeFn: (data: T) => U,
   options?: PaginationOptions,
-): Promise<List<U>> {
-  const finalOptions = setDefaultOptions(options);
-
+): Promise<List<U>> => {
   const { data } = await workos.get<ListResponse<T>>(endpoint, {
-    query: finalOptions,
+    query: setDefaultOptions(options),
   });
 
   return deserializeList(data, deserializeFn);
-}
+};
--- a/src/common/utils/pagination.ts
+++ b/src/common/utils/pagination.ts
@@ -1,4 +1,27 @@
-import { List, PaginationOptions } from '../interfaces';
+import { WorkOS } from '../../workos';
+import { List, ListResponse, PaginationOptions } from '../interfaces';
+import { deserializeList } from '../serializers';
+
+export function SetDefaultOptions(options?: PaginationOptions): PaginationOptions {
+  return {
+    ...options,
+    order: options?.order || 'desc',
+  };
+}
+
+export async function FetchAndDeserialize<T, U>(
+  workos: WorkOS,
+  endpoint: string,
+  deserializeFn: (data: T) => U,
+  options: PaginationOptions,
+): Promise<List<U>> {
+  const { data } = await workos.get<ListResponse<T>>(endpoint, {
+    query: options,
+  });
+
+  return deserializeList(data, deserializeFn);
+}
+
 
 export class AutoPaginatable<T> {
   readonly object: 'list' = 'list';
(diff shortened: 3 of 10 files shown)


src/common/interfaces/list.interface.ts
```diff
@@ -5,6 +5,10 @@ export interface List<T> {
     before?: string;
     after?: string;
   };
+  params?: {
```
maxdeviant, line 8: It doesn't seem like this belongs on the `List`, since it's not a value returned from the API.

src/common/interfaces/index.ts
```diff
@@ -5,3 +5,4 @@ export * from './put-options.interface';
 export * from './unprocessable-entity-error.interface';
 export * from './workos-options.interface';
 export * from './workos-response-error.interface';
+export * from './pagination-options.interface'
```
willmanduffy, line 8: You're going to want to make sure your code editor is setup to "Add new line at end of file" on save.

src/common/utils/pagination.ts
```diff
@@ -0,0 +1,41 @@
+import { DeserializedList, PaginationOptions} from '../interfaces';
+// Autopaginatable class definition
```
willmanduffy, line 2: We don't need this comment. It's basically just a rephrasing of the next line :)

src/common/utils/pagination.ts
```diff
@@ -0,0 +1,41 @@
...
+export class Autopaginatable<T> {
+  constructor(private _list: DeserializedList<T>, private apiCall: (params: PaginationOptions) => Promise<Autopaginatable<T>>) {}
```
willmanduffy, line 4: Your code editor might not be setup to run Prettier. (You can also do a `yarn format`).

src/common/utils/pagination.ts
```diff
@@ -0,0 +1,41 @@
+import { DeserializedList, PaginationOptions} from '../interfaces';
+// Autopaginatable class definition
+export class Autopaginatable<T> {
```
willmanduffy, line 3: I feel this should be `AutoPaginatable`. Especially since the function is `autoPagination` later.

src/common/interfaces/list.interface.ts
AshGodfrey: I think this is the crux of auto pagination - this is how Python is currently handling it (adding it to the object). In Node, what's the best way to handle this?

src/common/interfaces/index.ts
AshGodfrey: awesome tip - done.

src/common/interfaces/list.interface.ts
maxdeviant: We'll need to store the state somewhere else; perhaps on the `AutoPaginatable` class?

We just need to make sure that it's somewhere that can be tracked separately per method call. Since we create a new `AutoPaginatable` instance for each call I think that should be okay.

src/common/utils/pagination.ts
```diff
@@ -0,0 +1,43 @@
+import { DeserializedList, PaginationOptions } from '../interfaces';
+
+export class AutoPaginatable<T> {
+  constructor(
+    private _list: DeserializedList<T>,
```
maxdeviant, line 5: We don't need the leading `_` to signify that this is private, since we mark it as such with the `private` keyword:

```suggestion
    private list: DeserializedList<T>,
```

Additionally, the `_` prefix in particular will interfere with TypeScript's usage checking, as a `_` prefix is used to indicate to the compiler that a value is intentionally unused.

src/common/interfaces/list.interface.ts
AshGodfrey: Makes sense - thank you!

AshGodfrey: My tests are definitely going to fail, but maybe somehting like what I have now?

src/directory-sync/directory-sync.ts
```diff
@@ -21,18 +22,24 @@ import {
...
       '/directories',
       {
         query: options,
       },
     );
 
-    return deserializeList(data, deserializeDirectory);
+    return new AutoPaginatable(
+      deserializeList(data, deserializeDirectory),
+      (params) => this.listDirectories(params),
```
maxdeviant, line 40: One thing we'll need to check is that `AutoPaginatable ` is stack-safe.

Right now it seems like `listDirectories` and `AutoPaginatable` are mutually recursive:
- `listDirectories` creates an instance of `AutoPaginatable` that holds a handle to `listDirectories`
- On each page we request we'll make a call to `listDirectories`, which will create a new `AutoPaginatable` instance and recurse again

With enough pages being paginated through this could lead to a stack overflow.

AshGodfrey: yup - exactly what is happening. 

@maxdeviant with that feedback in mind I changed it to return a `DeserializedList` instead of `AutoPaginatable` - I think that solves it but I'm afraid I've added more chaos to get there.

src/common/utils/fetch-and-deserialize.ts
```diff
@@ -0,0 +1,25 @@
...
+  };
+}
+
+export async function fetchAndDeserialize<T, U>(
+  workos: WorkOS,
+  endpoint: string,
+  deserializeFn: (data: T) => U,
+  options?: PaginationOptions,
+): Promise<List<U>> {
+  const finalOptions = setDefaultOptions(options);
```
willmanduffy, line 18: Just a styling bit, I don't think we need to set this in another variable. Could just be `query: setDefaultOptions(options)` on line 21.

src/common/utils/fetch-and-deserialize.ts
```diff
@@ -0,0 +1,25 @@
...
+import { deserializeList } from '../serializers';
+
+function setDefaultOptions(options?: PaginationOptions): PaginationOptions {
+  return {
+    ...options,
+    order: options?.order || 'desc',
+  };
+}
+
+export async function fetchAndDeserialize<T, U>(
```
willmanduffy, line 12: Across the rest of the library we do `export const fetchAndDeserialize = async ()` and `const setDefaultOptions` so probably best to be consistent.

src/common/utils/pagination.ts
```diff
@@ -0,0 +1,55 @@
...
+export class AutoPaginatable<T> {
+  readonly object: 'list' = 'list';
+  readonly options: PaginationOptions;
+
+  constructor(
+    private list: List<T>,
+    private apiCall: (params: PaginationOptions) => Promise<List<T>>,
+    options?: PaginationOptions,
+  ) {
+    this.options = {
```
willmanduffy, line 12: Is this any different from `this.options = options`?

mthadley: Seems worth simplifying. We can always change it back later.

src/common/utils/pagination.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  get data(): T[] {
+    return this.list.data;
+  }
+
+  get listMetadata() {
+    return this.list.listMetadata;
+  }
+
+  private async *generatePages(params: PaginationOptions): AsyncGenerator<T[]> {
```
mthadley, line 25: Commenting for posterity: we should look into exposing this via [`Symbol.asyncIterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator) if later developers decide they'd like to lazily iterate over the pages directly.

Until then, only exposing `autoPagination` seems reasonable.

src/common/utils/pagination.ts
```diff
@@ -0,0 +1,55 @@
...
+    const result = await this.apiCall({
+      ...this.options,
+      after: params.after,
+    });
+
+    yield result.data;
+
+    if (result.listMetadata.after) {
+      yield* this.generatePages({ after: result.listMetadata.after });
```
mthadley, line 34: Related to [Marshall's previous comment](https://github.com/workos/workos-node/pull/788#discussion_r1288869625): Is this also a recursive call that would impact stack safety? If so, we could refactor this to use to use a `while` or `do while`.

I'm not sure how much it matters in practice though, so just noting it.

src/common/utils/pagination.ts
AshGodfrey: I get an undefined error if I don't use the spread operator, not sure if there's another way around that I should be taking.

mthadley (approved):

PR #790, opened 2023-08-04
Use `/users/sessions/token` and `password` grant type

mthadley:
## Description

`/users/sessions/token?grant_type=password` is the endpoint for password-based sessions going forward.

## Documentation

```
[ ] Yes
```


maxdeviant:
I think we're close on landing the renaming of the User Management APIs to camelCase, so we might want to wait for that to land first.

mthadley:
> I think we're close on landing the renaming of the User Management APIs to camelCase, so we might want to wait for that to land first.

Go ahead and get that merged. I'll update this one to be compatible with those changes.

cmatheson (approved):

maxdeviant:
@mthadley Looks like #791 just landed, so this PR can be rebased now.

src/users/users.ts
```diff
@@ -102,7 +105,6 @@ export class Users {
       serializeAuthenticateUserWithTokenOptions({
         ...payload,
         clientSecret: this.workos.key,
-        grantType: 'authorization_code',
```
mthadley, line 105: I moved the `grantType` down into the serializer since it never varies, and also updated the types to be the associated string literal rather than `string`.

src/users/serializers/authenticate-user-with-password-options.serializer.ts
```diff
@@ -1,15 +1,19 @@
...
+    AuthenticateUserWithPasswordCredentials,
 ): SerializedAuthenticateUserWithPasswordOptions => ({
+  grant_type: 'password',
+  client_id: options.clientId,
+  client_secret: options.clientSecret,
   email: options.email,
   password: options.password,
   ip_address: options.ipAddress,
   user_agent: options.userAgent,
-  start_session: options.startSession,
```
mthadley, line 13: Neither of these `/users/sessoins/token` API's take `start_session` anymore.

PR #793, opened 2023-08-07
Migrate SSO to camelCased

willmanduffy:
## Description

* Part of a wider initiative to migrate our Node SDK to camel case for the 3.0.0 release.
* Migrates the Single Sign-On module to camel case.

## Documentation

```
[ ] Yes
```


src/sso/interfaces/connection.interface.ts
```diff
@@ -7,6 +7,22 @@ export interface ConnectionDomain {
 }
 
 export interface Connection {
+  object: 'connection';
+  id: string;
+  organizationId?: string;
+  name: string;
+  connectionType: ConnectionType;
```
maxdeviant, line 14: What do we think about renaming this to `type`?

Now that we have the deserialization layer in place, we can translate the `connection_type` field to `type`.

This would align with `Directory`s, which already use `type`.

willmanduffy: I'm going to take that approach! I think that feels measured.

maxdeviant (approved):

PR #794, opened 2023-08-07
Add update user password method

danerwilliams:
## Description

Adds the update user password method for the user management API

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/users/users.ts
+++ b/src/users/users.ts
@@ -46,6 +46,7 @@ import {
   serializeCreatePasswordResetChallengeOptions,
   serializeCreateUserOptions,
   serializeRevokeSessionOptions,
+  serializeUpdateUserPassword,
   serializeVerifySessionOptions,
 } from './serializers';
 import { deserializeList } from '../common/serializers';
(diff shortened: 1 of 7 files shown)


src/users/users.ts
```diff
@@ -217,4 +219,18 @@ export class Users {
...
   }
+
+  async updateUserPassword({
+    userId,
+    password,
+  }: UpdateUserPasswordOptions): Promise<User> {
+    const { data } = await this.workos.put<
+      UserResponse,
+      SerializedUpdateUserPasswordOptions
+    >(`/users/${userId}/password`, {
```
maxdeviant, line 230: Should we have a dedicated serialization function like we do here? https://github.com/workos/workos-node/blob/c99d90819c6241d3612f18f42aaf62195f877a58/src/organizations/organizations.ts#L40

src/users/users.ts
```diff
@@ -217,4 +220,16 @@ export class Users {
 
     return deserializeUser(data);
   }
+
+  async updateUserPassword(payload: UpdateUserPasswordOptions): Promise<User> {
+    const { data } = await this.workos.put<
+      UserResponse,
+      SerializedUpdateUserPasswordOptions
```
maxdeviant, line 227: Based on the other usage, it looks like we might be able to infer this second type parameter:

```suggestion
```

maxdeviant (approved):

PR #795, opened 2023-08-07
Add update user method

danerwilliams:
## Description

Adds an `updateUser` method for user management

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/users/users.ts
+++ b/src/users/users.ts
@@ -25,6 +25,7 @@ import {
   SerializedCreateUserOptions,
   SerializedRevokeSessionOptions,
   SerializedVerifySessionOptions,
+  UpdateUserOptions,
   UpdateUserPasswordOptions,
   User,
   UserResponse,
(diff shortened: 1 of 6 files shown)


src/users/users.ts
```diff
@@ -217,4 +219,20 @@ export class Users {
...
+
+  async updateUser({
+    userId,
+    lastName,
+    firstName,
+  }: UpdateUserOptions): Promise<User> {
+    const { data } = await this.workos.put<
+      UserResponse,
+      SerializedUpdateUserOptions
+    >(`/users/${userId}`, {
```
maxdeviant, line 231: Should we have a dedicated serialization function like we do here? https://github.com/workos/workos-node/blob/c99d90819c6241d3612f18f42aaf62195f877a58/src/organizations/organizations.ts#L40

maxdeviant (approved):

PR #801, opened 2023-08-08
Consolidate Webhook and Event types in to Event

willmanduffy:
## Description

* Despite being identical payloads from the API, we had previously created separate Webhook and Event types with very similar type signatures. The key difference was that the Event data was not strongly typed to the response we would get from the API.
* This change migrates to a single Event type that has a more specific response signature.
* Additionally, we moved the webhook specific serializers into their respective domains instead of bundling them under webhook serializers. This makes them more re-usable from a common location.
* Fixes USRLD-583, USRLD-589

## Documentation

```
[ ] Yes
```


src/common/interfaces/event.interface.ts
```diff
@@ -0,0 +1,223 @@
...
+  DirectoryUser,
+  DirectoryUserResponse,
+  EventDirectory,
+  EventDirectoryGroup,
+  EventDirectoryGroupResponse,
+  EventDirectoryResponse,
+} from '../../directory-sync/interfaces';
+import { Connection, ConnectionResponse } from '../../sso/interfaces';
+
+export interface EventBase {
```
willmanduffy, line 11: This file is largely just the existing `Webhook` types renamed to `Event` and moved to the common folder.

src/common/interfaces/event.interface.ts
```diff
@@ -0,0 +1,223 @@
...
+  | DsyncGroupCreatedEventResponse
+  | DsyncGroupUpdatedEventResponse
+  | DsyncGroupDeletedEventResponse
+  | DsyncGroupUserAddedEventResponse
+  | DsyncGroupUserRemovedEventResponse
+  | DsyncUserCreatedEventResponse
+  | DsyncUserUpdatedEventResponse
+  | DsyncUserDeletedEventResponse;
+
+export type EventNames =
```
willmanduffy, line 209: This type existed for the Event API, but not webhooks previously, but as I combined they I took this over from Events.

src/directory-sync/interfaces/directory-group.interface.ts
```diff
@@ -19,3 +19,25 @@ export interface DirectoryGroupResponse {
   updated_at: string;
   raw_attributes: any;
 }
+
+export interface EventDirectoryGroup<TRawAttributes = any> {
```
willmanduffy, line 23: This was the existing `WebhookDirectoryGroup` type, but I took it back into the `directory-sync` folder since it didn't make sense for this to live in the `common` folder.

src/directory-sync/interfaces/directory.interface.ts
```diff
@@ -1,3 +1,24 @@
+export type DirectoryType =
```
willmanduffy, line 1: Now that `DirectoryType` exists in this file I'm going to come back later in https://linear.app/workos/issue/USRLD-632/change-directorytype-to-directorytype-instead-of-string to update line 30 to point towards `DirectoryType` instead of `string` as well. 

I just want to make sure it's in a separate OR so it doesn't get lost and we include it in the changelog.

src/common/interfaces/event.interface.ts
maxdeviant, line 209: Can we name this in the singular?

```suggestion
export type EventName =
```

Union types should use singular naming since they only can contain a single inhabitant value at a time.

src/events/events.ts
```diff
@@ -1,9 +1,7 @@
 import { WorkOS } from '../workos';
-import { Event, EventResponse } from './interfaces';
 import { ListEventOptions } from './interfaces';
-import { deserializeList } from '../common/serializers';
-import { List, ListResponse } from '../common/interfaces';
-import { deserializeEvent } from './serializers/event.serializer';
+import { deserializeEvent, deserializeList } from '../common/serializers';
```
willmanduffy, line 3: This change is important. Just by swapping the import location of the `deserializeEvent` we're now swapping out for the `deserializeEvent` that I wrote for webhooks. This fixes https://linear.app/workos/issue/USRLD-589/deserialize-event-responses

src/webhooks/serializers/index.ts
```diff
@@ -1,3 +0,0 @@
-export * from './webhook-directory-group.serializer';
```
willmanduffy, line 1: As stated elsewhere, all of these were pulled in to `directory-group.serializer`, `directory-user.serializer` and `directory.serializer` respectively.

willmanduffy:
Fixes USRLD-589

src/common/interfaces/event.interface.ts
willmanduffy: Sure can, copypasta :)

maxdeviant (approved):

PR #803, opened 2023-08-08
Fix `updateUser` options

danerwilliams:
## Description

`updateUser` options should require at least one property, not all properties to be passed. 

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/users/serializers/update-user-options.serializer.ts
+++ b/src/users/serializers/update-user-options.serializer.ts
@@ -2,7 +2,8 @@ import { SerializedUpdateUserOptions, UpdateUserOptions } from '../interfaces';
 
 export const serializeUpdateUserOptions = (
   options: UpdateUserOptions,
-): SerializedUpdateUserOptions => ({
-  first_name: options.firstName,
-  last_name: options.lastName,
-});
+): SerializedUpdateUserOptions =>
+  ({
+    first_name: options.firstName,
+    last_name: options.lastName,
+  } as SerializedUpdateUserOptions);
--- a/src/users/serializers/update-user-options.serializer.ts
+++ b/src/users/serializers/update-user-options.serializer.ts
@@ -3,12 +3,12 @@ import { SerializedUpdateUserOptions, UpdateUserOptions } from '../interfaces';
 export const serializeUpdateUserOptions = (
   options: UpdateUserOptions,
 ): SerializedUpdateUserOptions => {
-  if (!Object.keys(options).length) {
-    throw new Error('At least one property must be provided');
+  if (!options.firstName || !options.lastName) {
+    throw new Error('At least one of the properties must be provided');
   }
 
   return {
     first_name: options.firstName,
     last_name: options.lastName,
-  } as SerializedUpdateUserOptions;
+  };
 };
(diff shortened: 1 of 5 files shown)


src/users/serializers/update-user-options.serializer.ts
```diff
@@ -2,7 +2,13 @@ import { SerializedUpdateUserOptions, UpdateUserOptions } from '../interfaces';
...
-});
+): SerializedUpdateUserOptions => {
+  if (Object.keys(options).length === 0) {
+    throw new Error('At least one property must be provided');
+  }
+
+  return {
+    first_name: options.firstName,
+    last_name: options.lastName,
+  } as SerializedUpdateUserOptions;
```
danerwilliams, line 13: I don't like using an `as` here, but it seems like the only way to get this to compile (besides maybe being exceptionally verbose).

Adding a check to throw an error if for some reason there weren't at least 1 property passed I think works just as well here.

maxdeviant: I think this is a case where we want to use [`satisfies`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator).

The SDK targets TypeScript 5, so we should have access to it.

src/users/serializers/update-user-options.serializer.ts
```diff
@@ -2,7 +2,13 @@ import { SerializedUpdateUserOptions, UpdateUserOptions } from '../interfaces';
 
 export const serializeUpdateUserOptions = (
   options: UpdateUserOptions,
-): SerializedUpdateUserOptions => ({
-  first_name: options.firstName,
-  last_name: options.lastName,
-});
+): SerializedUpdateUserOptions => {
+  if (Object.keys(options).length === 0) {
```
maxdeviant, line 6: nit: More idiomatic to write the length check like this:

```suggestion
  if (!Object.keys(options).length) {
```

maxdeviant (approved):

src/users/serializers/update-user-options.serializer.ts
danerwilliams: Wasn't able to get `satisfies` to help with this. 

I thought I had tried this but I can just list out all of the properties in a check though to compile:
```ts
export const serializeUpdateUserOptions = (
  options: UpdateUserOptions,
): SerializedUpdateUserOptions => {
  if (!options.firstName || !options.lastName) {
    throw new Error('At least one of the properties must be provided');
  }

  return {
    first_name: options.firstName,
    last_name: options.lastName,
  };
};
```

kinda annoying but I think it's better than using an `as` with the generic check for empty object

danerwilliams: oh wait lol this doesn't make logical sense that's why I didn't do it

danerwilliams: need my brain to wake up haha

danerwilliams: I think we need to just stick with the `as`:

```ts
export const serializeUpdateUserOptions = (
  options: UpdateUserOptions,
): SerializedUpdateUserOptions => {
  if (!Object.keys(options).length) {
    throw new Error('At least one of the properties must be provided');
  }

  return {
    first_name: options.firstName,
    last_name: options.lastName,
  } as SerializedUpdateUserOptions;
};
```

`satisfies` won't narrow down the type down any further in this case

PR #804, opened 2023-08-09
Add workos.users.authenticateUserWithMagicAuth

willmanduffy:
## Documentation

```
[ ] Yes
```


src/users/users.ts
```diff
@@ -82,6 +85,24 @@ export class Users {
     return deserializeUser(data);
   }
 
+  async authenticateUserWithMagicAuth(
+    payload: AuthenticateUserWithMagicAuthOptions,
+  ): Promise<AuthenticationResponse> {
+    const { data } = await this.workos.post<
+      AuthenticationResponseResponse,
```
robframpton, line 92: Apologies if that has been discussed elsewhere already, but have we thought about alternative names for this interface?

willmanduffy: I'd prefer to keep it consistent across the board. We have `Entity` and `EntityResponse`. It's not ideal, but I think being inconsistent would be worse.

robframpton (approved):

PR #811, opened 2023-08-09
Make `session` required on `AuthenticationResponse`

maxdeviant:
## Description

This PR fixes a bug in the `AuthenticationResponse` type where the `session` field was marked as possibly being `undefined`.

The `session` will always be present on the `AuthenticationResponse`.

## Documentation

```
[ ] Yes
```


willmanduffy:
In one case it appears to be marked as optional in the Docs. Is this where this came from? 

CleanShot 2023-08-09 at 13 41 35@2x

maxdeviant:
> In one case it appears to be marked as optional in the Docs. Is this where this came from?
> 
> CleanShot 2023-08-09 at 13 41 35@2x

It was optional at one point (contingent on the value of `start_session`).

But those docs are out of date now, since `start_session` has been removed and we now always return the `session`.

maxdeviant:
@oliverzheng Can you confirm the above? ☝🏻

oliverzheng:
Can confirm. The doc is indeed out of date. Instead of removing the `start_session` param, the `sessions/token` endpoint now supports password auth and has obviated the `sessions/authentications` endpoint which supports the `start_session` param. The SDK has already been updated to use the new endpoint; the docs should be updated to reflect the new endpoint as well and the old endpoint should be removed entirely.

PR #813, opened 2023-08-10
Update revoke all sessions for user to use an interface

AshGodfrey:
## Description

Docs PR: https://github.com/workos/workos/pull/22357

## Documentation

```
[x ] Yes
```

### DIFF
--- a/src/users/interfaces/revoke-all-sessions-for-user-options.interface.ts
+++ b/src/users/interfaces/revoke-all-sessions-for-user-options.interface.ts
@@ -1,0 +1,3 @@
+export interface RevokeAllSessionsForUserOptions {
+    id: string;
+  }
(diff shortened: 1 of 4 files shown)


src/users/interfaces/revoke-all-sessions-for-user-options.interface.ts
```diff
@@ -0,0 +1,3 @@
+export interface RevokeAllSessionsForUserOptions {
+  id: string;
```
willmanduffy, line 2: I think we should call this `userId`

AshGodfrey: implementing + updating docs accordingly

willmanduffy (approved):

AshGodfrey:
@willmanduffy I can't merge into the node SDK, mind doing that for me?

PR #822, opened 2023-08-16
Fix: Properly serialize args for listConnections and listDirectories

willmanduffy:
## Description

* When we added serialization, we introduced a bug for these two endpoints that did not properly pass through the serialization of arguments for these two endpoints. This resolves the issue and adds tests to check for regressions in the future.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/directory-sync/interfaces/list-directories-options.interface.ts
+++ b/src/directory-sync/interfaces/list-directories-options.interface.ts
@@ -5,3 +5,9 @@ export interface ListDirectoriesOptions extends PaginationOptions {
   organizationId?: string;
   search?: string;
 }
+
+export interface SerializedListDirectoriesOptions extends PaginationOptions {
+  domain?: string;
+  organization_id?: string;
+  search?: string;
+}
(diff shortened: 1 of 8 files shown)


src/directory-sync/interfaces/list-directories-options.interface.ts
```diff
@@ -5,3 +5,9 @@ export interface ListDirectoriesOptions extends PaginationOptions {
   organizationId?: string;
   search?: string;
 }
+
+export interface SerializedListDirectoriesOptions extends PaginationOptions {
```
maxdeviant, line 9: Should we have a separate `SerializedPaginationOptions`?

willmanduffy: It's not really serializing them since there's no diff. I found that to be pretty weird?

maxdeviant: What if we were to add a new field to `PaginationOptions` in the future that did need to be serialized?

willmanduffy: So your take is that we should have a no-op serialization for all pagination endpoints in case we add this in the future? I can see this as a safeguard though I'm not extremely driven to add it personally. If you'd feel safer with it though happy to add.

PR #824, opened 2023-08-17
Add `enrollUserInMfaFactor` method

danerwilliams:
## Documentation

```
[ ] Yes
```


src/users/interfaces/enroll-user-in-mfa-factor.interface.ts
```diff
@@ -0,0 +1,12 @@
+export interface EnrollUserInMfaFactorOptions {
+  userId: string;
+  type: 'totp';
```
brockspratlen, line 3: WebAuthn is going to be a pretty quick add (next few weeks), so maybe we shouldn't hardcode `totp` here and instead have folks pass in the type?

brockspratlen (approved):

PR #826, opened 2023-08-18
Update group types for DsyncGroupUser events

awolfden:
## Description
Update group types to return full group object for DsyncGroupUser events

## Documentation

```
[ ] Yes
```


src/common/interfaces/event.interface.ts
```diff
@@ -113,7 +113,7 @@ export interface DsyncGroupUserAddedEvent extends EventBase {
   data: {
     directoryId: string;
     user: DirectoryUser;
-    group: Pick<EventDirectoryGroup, 'id' | 'name'>;
+    group: EventDirectoryGroupResponse;
```
willmanduffy, line 116: This should be an `EventDirectoryGroup`

src/common/interfaces/event.interface.ts
```diff
@@ -131,7 +131,7 @@ export interface DsyncGroupUserRemovedEvent extends EventBase {
...
```
willmanduffy, line 134: This should be an `EventDirectoryGroup`

src/common/interfaces/event.interface.ts
awolfden: Yep, good call - initially these were both throwing errors with EventDirectoryGroupResponse, but I just needed to adjust the event.serializer.ts file a bit to make it work.

src/common/serializers/event.serializer.ts
```diff
@@ -58,7 +58,7 @@ export const deserializeEvent = (event: EventResponse): Event => {
         data: {
           directoryId: event.data.directory_id,
           user: deserializeDirectoryUser(event.data.user),
-          group: event.data.group,
+          group: deserializeEventDirectoryGroup(event.data.group),
```
willmanduffy, line 61: Actually looks like `EventDirectoryGroup` and `DirectoryGroup` are now identical. Could we make that full migration and also make this `deserializeDirectoryGroup`?

Sorry missed this before

package-lock.json
```diff
@@ -0,0 +1,3167 @@
+{
```
willmanduffy, line 1: We won't want to commit this

awolfden: 🤦🏽

src/common/serializers/event.serializer.ts
awolfden: ✅

src/common/interfaces/event.interface.ts
```diff
@@ -122,7 +122,7 @@ export interface DsyncGroupUserAddedEventResponse extends EventResponseBase {
   data: {
     directory_id: string;
     user: DirectoryUserResponse;
-    group: Pick<EventDirectoryGroupResponse, 'id' | 'name'>;
+    group: EventDirectoryGroupResponse;
```
willmanduffy, line 125: Is `EventDirectoryGroupResponse` the same as `DirectoryGroupResponse` at this point?

awolfden: Sure is! Removed the event version in favor of DirectoryGroupResponse.

src/common/serializers/event.serializer.ts
```diff
@@ -58,7 +58,7 @@ export const deserializeEvent = (event: EventResponse): Event => {
         data: {
           directoryId: event.data.directory_id,
           user: deserializeDirectoryUser(event.data.user),
-          group: event.data.group,
+          group: deserializeDirectoryGroup(event.data.group),
```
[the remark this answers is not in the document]
awolfden: We can and we will 💪🏽

willmanduffy (approved):
Looks like there's a formatting problem

PR #837, opened 2023-08-24
Migrate completeEmailVerification to verifyEmail

willmanduffy:
## Description

* Changes the completeEmailVerification path and body to match the latest updates from the API. It now not only takes a token (now renamed to code) as an argument, but also requires the userId of the user.
* Fixes USRLD-849, USRLD-934

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/users/users.spec.ts
+++ b/src/users/users.spec.ts
@@ -193,15 +193,18 @@ describe('UserManagement', () => {
       });
     });
 
-    describe('completeEmailVerification', () => {
+    describe('verifyEmail', () => {
       it('sends a Complete Email Verification request', async () => {
-        mock.onPost(`/users/email_verification`).reply(200, userFixture);
+        mock.onPost(`/users/user-123/verify_email`).reply(200, userFixture);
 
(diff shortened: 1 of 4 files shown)


src/users/users.spec.ts
```diff
@@ -193,15 +193,18 @@ describe('UserManagement', () => {
...
+    describe('verifyEmail', () => {
       it('sends a Complete Email Verification request', async () => {
-        mock.onPost(`/users/email_verification`).reply(200, userFixture);
+        mock.onPost(`/users/user-123/verify_email`).reply(200, userFixture);
 
-        const resp = await workos.users.completeEmailVerification(
-          'email-verification-token',
-        );
+        const resp = await workos.users.verifyEmail({
+          userId: 'user-123',
```
maxdeviant, line 201: nit: Can we make this match the structure of the IDs we'll see in the wild?

```suggestion
          userId: 'user_123',
```

maxdeviant (approved):

danerwilliams (approved):

PR #843, opened 2023-08-25
Update `sendMagicAuthCode`

danerwilliams:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/users/users.spec.ts
+++ b/src/users/users.spec.ts
@@ -222,11 +222,11 @@ describe('UserManagement', () => {
           user: {
             object: 'user',
             email: 'test01@example.com',
-            firstName: 'Test 01',
-            lastName: 'User',
-            emailVerified: true,
-            createdAt: '2023-07-18T02:07:19.911Z',
-            updatedAt: '2023-07-18T02:07:19.911Z',
+            first_name: 'Test 01',
+            last_name: 'User',
+            email_verified: true,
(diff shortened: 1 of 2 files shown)


maxdeviant (dismissed):

src/users/users.spec.ts
```diff
@@ -219,7 +219,15 @@ describe('UserManagement', () => {
           email_address: 'bob.loblaw@example.com',
         })
         .reply(200, {
-          id: 'auth_challenge_01E4ZCR3C56J083X43JQXF3JK5',
+          user: {
+            object: 'user',
+            email: 'test01@example.com',
+            firstName: 'Test 01',
+            lastName: 'User',
+            emailVerified: true,
```
maxdeviant, line 227: The shape of this response isn't correct, as the API will return snake_cased values:

```suggestion
            email_verified: true,
```

src/users/users.ts
```diff
@@ -163,9 +162,9 @@ export class Users {
 
   async sendMagicAuthCode(
     options: SendMagicAuthCodeOptions,
-  ): Promise<MagicAuthChallenge> {
+  ): Promise<{ user: User }> {
     const { data } = await this.workos.post<
-      MagicAuthChallenge,
+      { user: User },
```
maxdeviant, line 167: We'll need to use the serialized representation here and then deserialize appropriately.

maxdeviant (approved):

PR #847, opened 2023-08-28
Add `rawData` to `OauthException`

maxdeviant:
## Description

This PR adds the `rawData` field containing the raw error response data to the `OauthException`s thrown by the SDK.

Resolves USRLD-1006.

## Documentation

```
[ ] Yes
```


src/common/exceptions/generic-server.exception.ts
```diff
@@ -1,18 +1,16 @@
 export class GenericServerException extends Error {
   readonly name: string = 'GenericServerException';
-  readonly rawData: unknown;
   readonly message: string = 'The request could not be completed.';
 
   constructor(
     readonly status: number,
     message: string | undefined,
-    rawData: unknown,
+    readonly rawData: unknown,
```
maxdeviant, line 8: Changed the `GenericServerException` to always set the `rawData`, as it doesn't make sense to only set it when a `message` is present.

danerwilliams (approved):

PR #848, opened 2023-08-28
Update user authentication methods to user `/users/authenticate` rather than `/users/sessions/token`

danerwilliams:
## Description

The primary changes here are just removing the `expires_in` parameter, switching over to the new `/users/authenticate` endpoint, and removing `session` from the responses

## Documentation

```
[ ] Yes
```


src/users/serializers/user-authentication.serializer.ts
```diff
@@ -1,22 +1,9 @@
...
-  expiresAt: session.expires_at,
-  authorizedOrganizations: session.authorized_organizations,
-  unauthorizedOrganizations: session.unauthorized_organizations.map(
-    deserializeUnauthorizedOrganization,
-  ),
-  token: session.token,
-});
+} from '../interfaces/user-authentication.interface';
 
 export const deserializeUnauthorizedOrganization = (
```
maxdeviant, line 8: Is this function being used anywhere?

danerwilliams: removed 🔪

maxdeviant (approved):

PR #880, opened 2023-11-09
Remove `domains` and `allowProfilesOutsideOrganization` from `Organization` entity

robframpton:
## Description

Breaking change. Remove `domains` and `allowProfilesOutsideOrganization` properties from the `Organization` entity.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/organizations/fixtures/create-organization.json
+++ b/src/organizations/fixtures/create-organization.json
@@ -2,12 +2,5 @@
   "name": "Test Organization",
   "object": "organization",
   "id": "org_01EHT88Z8J8795GZNQ4ZP1J81T",
-  "allow_profiles_outside_organization": false,
-  "domains": [
-    {
-      "domain": "example.com",
-      "object": "organization_domain",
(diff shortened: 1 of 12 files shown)


src/organizations/fixtures/create-organization.json
```diff
@@ -2,12 +2,5 @@
...
   "id": "org_01EHT88Z8J8795GZNQ4ZP1J81T",
-  "allow_profiles_outside_organization": false,
-  "domains": [
-    {
-      "domain": "example.com",
-      "object": "organization_domain",
-      "id": "org_domain_01EHT88Z8WZEFWYPM6EC9BX2R8"
-    }
-  ]
+  "allow_profiles_outside_organization": false
```
LauraBeatris, line 5: Why does `allow_profiles_outside_organization`  remains in the JSON fixture if we're removing it?

robframpton: It still technically comes back from the API, but it can be removed safely as well.

LauraBeatris: Oh got it, so it'll remain the schema.

PR #881, opened 2023-11-10
Edit directory user interface to reflect nullable attributes

blairlunceford:
## Description

**BREAKING CHANGE - PUT IN MAJOR RELEASE**

Fixes the directory user interface to accurately reflect the typing of attributes in the API.

## Documentation

```
[x] Yes
```

Docs PR: https://github.com/workos/workos/pull/23767


src/directory-sync/interfaces/directory-user.interface.ts
```diff
@@ -37,19 +37,19 @@ export interface DirectoryUserResponse<
   object: 'directory_user';
   id: string;
   directory_id: string;
-  organization_id: string | null;
+  organization_id?: string;
   raw_attributes: TRawAttributes;
   custom_attributes: TCustomAttributes;
   idp_id: string;
-  first_name: string;
+  first_name?: string;
```
lukemorales, line 44: Since `undefined` is removed from JSON, a more accurate definition of what we expose would be for the property to exist, but be `string | null` since we get the value or fallback to `null` if it doesn't exist in our API.

```suggestion
  first_name: string | null;
```

Can you make this change for all properties you altered here?

blairlunceford: So I decided to change to the `?` pattern because that is more widely used in this SDK. Some examples:
https://github.com/workos/workos-node/blob/main/src/sso/interfaces/profile.interface.ts
https://github.com/workos/workos-node/blob/main/src/audit-logs/interfaces/audit-log-export.interface.ts

blairlunceford: Note: Luke and I discussed this in Slack, and I've added changes to have the interface reflect the data we currently emit from the API.

As a follow-up, we'll need to look at all interfaces in the SDK to ensure they match the data we actually send in the API.

lukemorales (approved):

PR #886, opened 2023-11-12
Remove `object` response property from Directory User response

LauraBeatris:
## Description

Remove `object` response property from Directory User response. It's not included in the API response, and also not listed in the [docs](https://workos.com/docs/reference/directory-sync/directory-user). 

CleanShot 2023-11-13 at 10 21 09

## Documentation

```
[ ] Yes
```


LauraBeatris:
The API just returns `directory_user` object key in the response.

PR #900, opened 2023-11-21
Add `authenticateWithEmailVerification` method

danerwilliams:
## Documentation

```
[ ] Yes
```

#### PR Dependency Tree

* **PR #900** 👈

### DIFF
--- a/src/user-management/interfaces/authenticate-with-email-verification-options.interface.ts
+++ b/src/user-management/interfaces/authenticate-with-email-verification-options.interface.ts
@@ -1,0 +1,21 @@
+export interface AuthenticateWithEmailVerificationOptions {
+  clientId: string;
+  code: string;
+  pendingAuthenticationToken: string;
+  ipAddress?: string;
+  userAgent?: string;
+}
+
+export interface AuthenticateUserWithEmailVerificationCredentials {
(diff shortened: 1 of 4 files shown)


benoitgrelard (commented):
How does this differ from `authenticateWithMagicAuth`?
That's the one I'm using currently for the magic auth example in OSAK.
It seems it's hitting the same endpoint, the only diff is no `userId` in `authenticateWithEmailVerification`.

danerwilliams:
> How does this differ from `authenticateWithMagicAuth`? That's the one I'm using currently for the magic auth example in OSAK. It seems it's hitting the same endpoint, the only diff is no `userId` in `authenticateWithEmailVerification`.

@benoitgrelard the main difference is the value set for the `grant_type`

jbvolta (approved):

benoitgrelard:
> > How does this differ from `authenticateWithMagicAuth`? That's the one I'm using currently for the magic auth example in OSAK. It seems it's hitting the same endpoint, the only diff is no `userId` in `authenticateWithEmailVerification`.
> 
> @benoitgrelard the main difference is the value set for the `grant_type`

When would you use one over the other?

PR #911, opened 2023-11-22
Fix `code` on `Challenge` which is incorrectly set as required

danerwilliams:
## Documentation

```
[ ] Yes
```

#### PR Dependency Tree

* **PR #911** 👈

### DIFF
--- a/src/mfa/interfaces/challenge.interface.ts
+++ b/src/mfa/interfaces/challenge.interface.ts
@@ -4,7 +4,7 @@ export interface Challenge {
   createdAt: string;
   updatedAt: string;
   expiresAt?: string;
-  code: string;
+  code?: string;
   authenticationFactorId: string;
 }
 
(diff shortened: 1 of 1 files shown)


jbvolta (approved):

src/mfa/interfaces/challenge.interface.ts
```diff
@@ -4,7 +4,7 @@ export interface Challenge {
   createdAt: string;
   updatedAt: string;
   expiresAt?: string;
-  code: string;
+  code?: string;
```
amygdalama, line 7: note that this interface is also shared w/ standalone MFA. does standalone have a required code?

amygdalama: oh it was made optional

amygdalama: 🤦‍♀️

amygdalama: ignore me!!!

PR #913, opened 2023-11-22
Make `sendPasswordResetEmail` response null

danerwilliams:
## Documentation

```
[ ] Yes
```

#### PR Dependency Tree

* **PR #913** 👈


src/user-management/user-management.spec.ts
```diff
@@ -360,12 +360,7 @@ describe('UserManagement', () => {
         `/user_management/password_reset/send`,
       );
 
-      expect(resp).toMatchObject({
-        token: 'password-reset-token',
-        user: {
-          email: 'test01@example.com',
-        },
-      });
+      expect(resp).toBeUndefined();
```
amygdalama, line 363: do we want undefined or an empty object? any existing examples in the SDK?

danerwilliams: we return void for delete methods, and the revokeInvitation method as an example.

I think it should be undefined, the response we receive from the API is no body, not an empty object

amygdalama (approved):

PR #927, opened 2023-11-27
Improve `Factor` typing

danerwilliams:
## Documentation

```
[ ] Yes
```

#### PR Dependency Tree

* **PR #927** 👈

### DIFF
--- a/src/mfa/interfaces/factor.interface.ts
+++ b/src/mfa/interfaces/factor.interface.ts
@@ -13,16 +13,6 @@ export interface Factor {
   totp?: Totp;
 }
 
-export interface UserManagementFactor {
-  object: 'authentication_factor';
-  id: string;
-  createdAt: string;
-  updatedAt: string;
-  type: 'totp';
-  totp: Totp;
-  userId: string;
-}
-
 export interface FactorResponse {
   object: 'authentication_factor';
   id: string;
@@ -32,14 +22,4 @@ export interface FactorResponse {
   sms?: SmsResponse;
   totp?: TotpResponse;
   user_id?: string;
-}
-
-export interface UserManagementFactorResponse {
-  object: 'authentication_factor';
-  id: string;
-  created_at: string;
-  updated_at: string;
-  type: 'totp';
-  totp: TotpResponse;
-  user_id: string;
 }
--- a/src/mfa/interfaces/factor.interface.ts
+++ b/src/mfa/interfaces/factor.interface.ts
@@ -11,7 +11,16 @@ export interface Factor {
   type: FactorType;
   sms?: Sms;
   totp?: Totp;
-  userId?: string;
+}
+
+export interface UserManagementFactor {
+  object: 'authentication_factor';
+  id: string;
+  createdAt: string;
+  updatedAt: string;
+  type: 'totp';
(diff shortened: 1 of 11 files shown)


src/mfa/interfaces/factor.interface.ts
```diff
@@ -11,7 +11,16 @@ export interface Factor {
   type: FactorType;
   sms?: Sms;
   totp?: Totp;
-  userId?: string;
+}
+
+export interface UserManagementFactor {
```
amygdalama, line 16: move to `user-management` ?

amygdalama: same w/ the deserializer etc

src/user-management/user-management.ts
```diff
@@ -48,9 +48,13 @@ import {
   Challenge,
   ChallengeResponse,
   Factor,
```
amygdalama, line 50: does this still need to use the MFA standalone Factor?

danerwilliams: ah yeah had some lingering usage, fixed now

src/mfa/interfaces/factor.interface.ts
danerwilliams: whoops actually created the files in wrong folder, fixed now

src/mfa/interfaces/totp.interface.ts
```diff
@@ -1,28 +1,15 @@
-interface TotpObject {
+export interface Totp {
   issuer: string;
   user: string;
-  qrCode: string;
-  secret: string;
-  uri: string;
+  qrCode?: string;
```
amygdalama, line 4: are these optional or required for standalone MFA responses?

danerwilliams: for both standalone and userland mfa, the secret properties (qr code, secret, and uri) are used in some cases and not used in others. 

so they should be optional for both

amygdalama: ah ok, so seems like the standalone MFA code for retrieving a totp factor was broken? 🤔  (asking for context so i know what to put in the changelog)

src/user-management/interfaces/user-management-factor.interface.ts
```diff
@@ -0,0 +1,21 @@
+import { Totp, TotpResponse } from '../../mfa/interfaces/totp.interface';
+
+export interface UserManagementFactor {
```
amygdalama, line 3: now that these are in `user-management`, what do you think about dropping the `UserManagement` prefix from the names?

src/mfa/interfaces/totp.interface.ts
danerwilliams: yes! the standalone mfa api was inacurrately saying that there was always access to these responses even in methods that didn't have them

src/user-management/interfaces/user-management-factor.interface.ts
danerwilliams: yeah that seems reasonable

danerwilliams: done

amygdalama (approved):

PR #962, opened 2024-02-09
Support Cloudflare workers/Edge runtime environments

benoitgrelard:
## Description

This PR aims to make the node SDK (will need to be renamed to JS SDK maybe?) compatible with Cloudflare workers / Edge runtime environments.

This is achieved by:
- Removing `axios` and using native `fetch` instead
- Removing usage of Node's `crypto` module in favour for native web `crypto`

Here are the important changes:
- Swapping out axios's client for [our own custom fetch-based client](https://github.com/workos/workos-node/pull/962/files#diff-32c7ee5e94c4dc53afedbcdb133765b9ebfeee5922e4b6a1fefd0987ee0d86d8) in [the main `WorkOS` class](https://github.com/workos/workos-node/pull/962/files#diff-43fae42284da1898f98b43b56a7bc6f5fb979e271bca59f4670751ba2c0020ef)
- Rewrite the webhooks APIs using web `crypto` in [the `Webhooks` class](https://github.com/workos/workos-node/pull/962/files#diff-8b772e2b2baa9169ed0a1e9039df9b4b4b353690593830c843dff21f3dc525f7)

Most of the rest of the changes are rewriting the tests to ensure they all pass because of the changes.

## Documentation

```
[x] Yes
```

> [!WARNING]  
> TODO: Add Docs PR


setup-jest.ts
```diff
@@ -0,0 +1,6 @@
+import { enableFetchMocks } from 'jest-fetch-mock';
+import { Crypto } from '@peculiar/webcrypto';
```
benoitgrelard, line 2: We need to add webcrypto support inside jest env.

src/common/interfaces/workos-options.interface.ts
```diff
@@ -1,8 +1,6 @@
-import { AxiosRequestConfig } from 'axios';
-
 export interface WorkOSOptions {
   apiHostname?: string;
   https?: boolean;
   port?: number;
-  axios?: Omit<AxiosRequestConfig, 'baseURL'>;
+  config?: RequestInit;
```
benoitgrelard, line 5: This is a breaking change as we were exposing `axios` specific options through the constructor.

src/common/utils/fetch-client.ts
benoitgrelard, line 1: This is the new fetch client that replaces `axios`'s.

src/common/utils/fetch-client.ts
```diff
@@ -0,0 +1,107 @@
...
+    return url.toString();
+  }
+
+  private async fetch(url: string, options?: RequestInit) {
+    const response = await fetch(url, {
+      ...this.options,
+      ...options,
+      headers: {
+        Accept: 'application/json, text/plain, */*',
+        'Content-Type': 'application/json',
```
benoitgrelard, line 76: default to `json` for all API calls.

src/common/utils/fetch-client.ts
```diff
@@ -0,0 +1,107 @@
...
+    path: string,
+    entity: Entity,
+    options: { params?: Record<string, any>; headers?: HeadersInit },
+  ) {
+    const resourceURL = this.getResourceURL(path, options.params);
+    const bodyIsSearchParams = entity instanceof URLSearchParams;
+    const contentTypeHeader = bodyIsSearchParams
+      ? { 'Content-Type': 'application/x-www-form-urlencoded;charset=utf-8' }
+      : undefined;
+    const body = bodyIsSearchParams ? entity : JSON.stringify(entity);
```
benoitgrelard, line 27: Support body as `URLSearchParams` for the one use-case we have for SSO. (`getProfileAndToken`)

src/common/utils/test-utils.ts
benoitgrelard, line 1: These are a bunch of utils I wrote so the tests were easier to rewrite and understand.

src/organizations/organizations.spec.ts
```diff
@@ -37,91 +42,75 @@ describe('Organizations', () => {
...
 
         const { data } = await workos.organizations.listOrganizations({
-          domains: ['example.com'],
+          domains: ['example.com', 'example2.com'],
         });
 
-        expect(mock.history.get[0].params).toEqual({
-          domains: ['example.com'],
+        expect(fetchSearchParams()).toEqual({
+          domains: 'example.com,example2.com',
```
benoitgrelard, line 52: This is one meaningful change here that I'd like to call out.

I have changed the output here because `axios` would serialize arrays automatically, and here we use `URLSearchParams` which can't do that, but do serialize using comma.

So with a single domain, I would end up with just `example.com` as a string. So I added a second one and updated the equal check so it's easier to understand what the SDK produces.

I have checked our API controllers and wherever we expect arrays, we do have a transform that works with both arrays and strings:

```tsx
@Transform(({ value }: { value: string | string[] }) => {
    if (Array.isArray(value)) {
      return value;
    }

    return value.split(',');
  })
```

I've also validated this all works using a local build of the new package against the Open Source AuthKit repo.

src/user-management/user-management.spec.ts
```diff
@@ -58,27 +60,27 @@ describe('UserManagement', () => {
...
         limit: 10,
       });
 
-      expect(mock.history.get[0].params).toEqual({
+      expect(fetchSearchParams()).toEqual({
         email: 'foo@example.com',
         organization_id: 'org_someorg',
         after: 'user_01H5JQDV7R7ATEYZDEG0W5PRYS',
-        limit: 10,
+        limit: '10',
```
benoitgrelard, line 75: Again, because we use `URLSearchParams`, everything is a string.

src/webhooks/webhooks.ts
```diff
@@ -1,10 +1,9 @@
-import crypto from 'crypto';
 import { SignatureVerificationException } from '../common/exceptions';
 import { deserializeEvent } from '../common/serializers';
 import { Event, EventResponse } from '../common/interfaces';
 
 export class Webhooks {
-  constructEvent({
+  async constructEvent({
```
benoitgrelard, line 6: This is another breaking change as pretty much all the methods are now async.

src/webhooks/webhooks.ts
```diff
@@ -72,37 +71,64 @@ export class Webhooks {
...
+    // encoded in hex to match the CryptoProvider contract. We map each byte in
+    // the buffer to its corresponding hex octet and then combine into a string.
+    const signatureBytes = new Uint8Array(signatureBuffer);
+    const signatureHexCodes = new Array(signatureBytes.length);
+
+    for (let i = 0; i < signatureBytes.length; i++) {
+      signatureHexCodes[i] = byteHexMapping[signatureBytes[i]];
+    }
+
+    return signatureHexCodes.join('');
```
benoitgrelard, line 106: Most of this was copied over from stripe: https://github.com/stripe/stripe-node/blob/master/src/crypto/SubtleCryptoProvider.ts

src/webhooks/webhooks.ts
```diff
@@ -72,37 +71,64 @@ export class Webhooks {
...
+    const bufferA = Buffer.from(stringA);
+    const bufferB = Buffer.from(stringB);
 
-    if (strA.length !== strB.length) {
+    if (bufferA.length !== bufferB.length) {
       return false;
     }
 
-    if (crypto.timingSafeEqual) {
-      return crypto.timingSafeEqual(strA, strB);
```
benoitgrelard, line 96: `timingSafeEqual` doesn't exist in web crypto so it led me to this issue and I used this double hmac method instead: https://github.com/w3c/webcrypto/issues/270#issuecomment-1899234835

src/workos.ts
benoitgrelard, line 1: In this file, everything should be mostly plug & play between the old `axios` client and the new fetch-based client.

tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "module": "commonjs",
     "declaration": true,
     "outDir": "lib",
-    "strict": true
+    "strict": true,
+    "lib": ["dom", "es2019"]
```
benoitgrelard, line 17: Needed to add some libs because I used `Object.entries` in my test utils.

package.json
```diff
@@ -34,15 +34,16 @@
...
   "dependencies": {
-    "axios": "~1.6.5",
     "pluralize": "8.0.0"
   },
   "devDependencies": {
+    "@peculiar/webcrypto": "^1.4.5",
     "@types/jest": "29.5.3",
     "@types/node": "14.18.54",
     "@types/pluralize": "0.0.30",
     "axios-mock-adapter": "1.21.5",
```
PaulAsjes, line 44: Do we still need this?

PaulAsjes (approved):
Looks good, just one tiny nit about a dependency.

package.json
benoitgrelard: Yes, see https://github.com/workos/workos-node/pull/962#discussion_r1484519580

PaulAsjes: I was referring to `"axios-mock-adapter": "1.21.5",`, looks like it's not used anywhere.

benoitgrelard: oh I see!

PR #965, opened 2024-02-19
Refactor Buffer usage

### LINKED ISSUE #964: Incompatibility with Cloudflare Workers
Although buffer is globally defined in a node environment, in an edge environment like Cloudflare Workers, Buffer has to be imported as "node:buffer". Frameworks like SvelteKit handle the conversion of "buffer" to "node:buffer" internally [Link](https://github.com/sveltejs/kit/pull/10544). However, in constructEvent, because Buffer is assumed to be globally defined. Screenshot 2024-02-17 225517 

When deployed to Cloudflare, my SvelteKit application's webhook errors out with "Buffer not defined"
Screenshot 2024-02-17 224746

benoitgrelard:
## Description

Fixes #964

There was a leftover usage of Node's `Buffer` API.
I have converted it to use the web platform's `TextEncoder` APIs instead.

I wonder if there's an ESLint plugin that could be used to lint for usages of APIs incompatible with cloudflare workers/edge runtime as it seems easy to let something through without thoroughly testing every single API call. A quick google search didn't bring much up so far.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/webhooks/webhooks.ts
+++ b/src/webhooks/webhooks.ts
@@ -3,6 +3,8 @@ import { deserializeEvent } from '../common/serializers';
 import { Event, EventResponse } from '../common/interfaces';
 
 export class Webhooks {
+  private encoder = new TextEncoder();
+
   async constructEvent({
     payload,
     sigHeader,
@@ -78,11 +80,10 @@ export class Webhooks {
   ): Promise<string> {
     payload = JSON.stringify(payload);
     const signedPayload = `${timestamp}.${payload}`;
-    const encoder = new TextEncoder();
 
     const key = await crypto.subtle.importKey(
       'raw',
-      encoder.encode(secret),
+      this.encoder.encode(secret),
       { name: 'HMAC', hash: 'SHA-256' },
       false,
       ['sign'],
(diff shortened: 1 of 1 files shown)


src/webhooks/webhooks.ts
```diff
@@ -107,8 +108,9 @@ export class Webhooks {
   }
 
   async secureCompare(stringA: string, stringB: string): Promise<boolean> {
-    const bufferA = Buffer.from(stringA);
-    const bufferB = Buffer.from(stringB);
+    const encoder = new TextEncoder();
```
PaulAsjes, line 111: Nit, but maybe this should be a class variable rather than being reinitialised.

PaulAsjes (approved):

src/webhooks/webhooks.ts
benoitgrelard: Good idea, done in [b0a8840](https://github.com/workos/workos-node/pull/965/commits/b0a8840ba59c1b57778f1e276db8eec889844e20)

PaulAsjes (approved):

PR #966, opened 2024-02-19
Allow null body

benoitgrelard:
## Description

This is follow-up from the issue raised in #963 

## Documentation

```
[ ] Yes
```


PaulAsjes (approved):
Not a blocker for this PR, but we should add tests for all the files in the `/utils` directory.

benoitgrelard:
> Not a blocker for this PR, but we should add tests for all the files in the `/utils` directory.

I've added a regression test in the workos test suite for now.

PaulAsjes (approved):

PR #974, opened 2024-02-29
Add session-management endpoints

cmatheson:
## Description

Adds types/endpoints for session management.

## Documentation

```
[X] Yes
```

### DIFF
--- a/src/user-management/user-management.ts
+++ b/src/user-management/user-management.ts
@@ -517,7 +517,7 @@ export class UserManagement {
     state,
   }: AuthorizationURLOptions): string {
     if (!provider && !connectionId && !organizationId) {
-      throw new Error(
+      throw new TypeError(
         `Incomplete arguments. Need to specify either a 'connectionId', 'organizationId', or 'provider'.`,
       );
     }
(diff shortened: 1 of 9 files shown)


src/user-management/user-management.ts
```diff
@@ -501,4 +536,12 @@ export class UserManagement {
 
     return `${this.workos.baseURL}/user_management/authorize?${query}`;
   }
+
+  getLogoutUrl({ sessionId }: { sessionId: string }): string {
+    if (!sessionId) {
+      throw new Error(`Incomplete arguments. Need to specify 'sessionId'.`);
```
mthadley, line 542: Should we throw a `TypeError` here instead?

I think it's generally the type used for errors in the way a function is called. ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/More_arguments_needed))

mthadley (approved):

src/user-management/user-management.ts
cmatheson: ah good call. i've changed this (and the error we were throwing in getAuthorizationUrl (which i copied from). i think that's probably safe to do? if someone was doing an instanceof check on the exception before that should still pass.

mthadley: > if someone was doing an instanceof check on the exception before that should still pass.

That's right! Both are instances of `Error`.

PR #977, opened 2024-03-04
v6.1.0

cmatheson:
## Description

This includes #974.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "6.0.3",
+  "version": "6.1.0",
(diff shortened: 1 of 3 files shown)


PaulAsjes (changes requested):
Since this adds backwards compatible functionality rather than bug fixes, I think this should be a minor version instead of a patch. Can we make it 6.1.0 instead?

cmatheson:
good call @PaulAsjes , fixed.

PaulAsjes (approved):

PR #978, opened 2024-03-09
Fix typeError on global.crypto assignment

Icegreeen:
## Description

This PR resolves an issue where a TypeError occurs when trying to assign the 'crypto' property to the global object in a Jest test environment. The error occurred due to trying to get a read-only property from the global object.

The original error:
`TypeError: Cannot assign read-only property 'crypto' of object '[global object]'`

Captura de tela 2024-03-09 062108

This was because the 'crypto' property already existed on the global object and was read-only.

To solve this problem I added an `if (!global.crypto)` check before assigning the 'crypto' property to the global object. This ensures that the 'crypto' property will only be set if it is not already set on the global object, thus avoiding the TypeError error.

Captura de tela 2024-03-09 062551

## Documentation

```
[ X ] No
```

### DIFF
--- a/setup-jest.ts
+++ b/setup-jest.ts
@@ -3,4 +3,6 @@ import { Crypto } from '@peculiar/webcrypto';
 
 enableFetchMocks();
 
-global.crypto = new Crypto();
+if (!global.crypto) {
+    global.crypto = new Crypto();
+}
--- a/setup-jest.ts
+++ b/setup-jest.ts
@@ -3,6 +3,7 @@ import { Crypto } from '@peculiar/webcrypto';
 
 enableFetchMocks();
 
+// Assign Node's Crypto to global.crypto if it is not already present
 if (!global.crypto) {
     global.crypto = new Crypto();
 }


setup-jest.ts
```diff
@@ -3,4 +3,6 @@ import { Crypto } from '@peculiar/webcrypto';
 
 enableFetchMocks();
 
-global.crypto = new Crypto();
+if (!global.crypto) {
```
LauraBeatris, line 6: Could we add a comment here to clarify this behavior? 

```suggestion
// Assign Node's Crypto to global.crypto if it is not already present
if (!global.crypto) {
```

PaulAsjes:
Thanks for the PR @Icegreeen!

Would you be able to provide some steps to reproduce this? I just ran all the tests on my machine on main and everything passed as expected.

Icegreeen:
> Thanks for the PR @Icegreeen!
> 
> Would you be able to provide some steps to reproduce this? I just ran all the tests on my machine on main and everything passed as expected.

I just tested in two environments: (Windows 11 and Linux mint) and the same error occurred.

- I cloned the project
- Executed `yarn`
- I ran `yarn test`

And the problem occurs

PaulAsjes (approved):
Thanks for this!

PR #984, opened 2024-03-13
v6.2.0

oliverzheng:
## Description

Includes #980 and #983.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "6.1.1",
+  "version": "6.2.0",
   "name": "@workos-inc/node",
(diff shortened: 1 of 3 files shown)


cmatheson (commented):
should this be 6.2.0 since we're adding new functionality?

oliverzheng:
Hmm I think you are right @cmatheson. As per https://semver.org/:

> MINOR version when you add functionality in a backward compatible manner
> PATCH version when you make backward compatible bug fixes

This is not a fix. I'll update.

cmatheson (approved):

PR #998, opened 2024-03-25
Add authentication events

LauraBeatris:
## Description

Adding types/parsing for authentication events - all the succeeded events and only the failed events for password and Magic Auth. We will add the other failed events as we complete support for them.

Resolves DAAP-913

## Documentation

```
[ ] Yes
```


src/user-management/interfaces/authentication-events.interface.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+  | {
+      status: 'failed';
+      error: {
+        code: string;
+        message: string;
+      };
+    };
+
+export type EmailVerification = {
```
LauraBeatris, line 13: I'm thinking of using this same module for other authentication event types, such as Magic Auth, MFA, etc, they share a similar shape, and create something like `magic-auth.interface.ts` looks like an entity when it's actually not 🤔

src/user-management/interfaces/authentication-events.interface.ts
```diff
@@ -0,0 +1,27 @@
...
+  object: 'email_verification';
+  email: string;
+  ipAddress: string | null;
+  userAgent: string | null;
+  userId: string | null;
+} & EventStatus;
```
LauraBeatris, line 19: This union will allow to have a precise type checking that `error` will always be defined when `status` is equal to `failed`.

src/user-management/interfaces/authentication-events.interface.ts
LauraBeatris: I could also create `email-verification.interface.ts` but with `EmailVerificationEvent` as the interface exported from there, I think this would be better.

amygdalama: what about one shared `AuthenticationEvent` type?

src/user-management/interfaces/authentication-events.interface.ts
```diff
@@ -0,0 +1,27 @@
...
+  | {
+      status: 'failed';
+      error: {
+        code: string;
+        message: string;
+      };
+    };
+
+export type EmailVerification = {
+  object: 'email_verification';
```
amygdalama, line 14: we have a `type` field but no `object` for these since they don't correspond to an API object.

src/user-management/interfaces/authentication-events.interface.ts
LauraBeatris: That sounds better to me! Going to update the code.

src/user-management/interfaces/authentication-events.interface.ts
amygdalama: as a start i'd recommend being looser with the types. we can always tighten them later if that seems useful. so what do you think about having one type where `error` is optional? it might be useful to take some inspiration from the types stripe defines in their sdk.

amygdalama: yep

LauraBeatris: Makes sense to me to leave it a bit looser, although I'd really love to get the exact states - with the above would mean that the error could be defined as `succeed` which is a bit weird but not a huge friction for the developer's app.

Let me look at how Stripe declares those types -- although I think they are auto-generated, right? So I would expect that it's a bit loose as well.

LauraBeatris: > although I think they are auto-generated

Which is what the future of this SDK will look like as well 😅

amygdalama: 🤞 lol

amygdalama: here's an example: https://github.com/stripe/stripe-node/blob/7293b69a204c4a25ca00574078f6b6699936b3cf/types/Charges.d.ts#L107

LauraBeatris: @amygdalama I don't think Stripe defines event errors on their API - from what I can see [here](https://github.com/stripe/stripe-node/blob/7293b69a204c4a25ca00574078f6b6699936b3cf/types/Events.d.ts#L305).

src/user-management/interfaces/authentication-event.interface.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export type AuthenticationEventType =
+  | 'email_verification'
+  | 'magic_auth'
+  | 'mfa'
+  | 'oauth'
+  | 'password'
+  | 'sso';
+
+export interface AuthenticationEvent<T extends AuthenticationEventType> {
```
LauraBeatris, line 14: Note for tomorrow: I think we should export separate interfaces per event type, this will allow us to expand for the case where individual ones have additional fields in the future. 

```ts
type AuthenticationPasswordEvent = AuthenticationEvent<'password'>
```

Could be expanded to:
```ts
type AuthenticationPasswordEvent = AuthenticationEvent<'password'> & { additionalField: 'true' }
```

amygdalama: i think for type-specific fields we can do something like what stripe does for their `PaymentMethod`: https://github.com/stripe/stripe-node/blob/7293b69a204c4a25ca00574078f6b6699936b3cf/types/PaymentMethods.d.ts#L25

LauraBeatris: Hmmm didn't understand, are they using namespaces to override/extend fields from the base type?

amygdalama: here's the API reference: https://docs.stripe.com/api/payment_methods/object

`PaymentMethod` objects have a `type` field with values like `card` or `affirm`. any data specific to the payment method type is then stored in a field named after the `type` value. so a `PaymentMethod` with a `type` of the value `card` will have a `card` field containing card-specific data.

the way this is implemented in the SDK is a `PaymentMethod` interface with optional fields corresponding to each of the `type` values. in the link above, `card` and `affirm` are optional fields on `PaymentMethod`s.

when we eventually have data specific to the authentication event `type`, we can use a similar pattern -- `sso`, `password`, etc can all be optional on the `AuthenticationEvent`.

the downside of the above approach: this doesn't allow the type system to guarantee that `sso`-type events will have a `sso` field. but i think that's ok for the use cases of these events (observability).

the benefit of the above approach is we have more flexibility to make non-breaking changes, which is useful especially because the failed authentication events are still a work in progress. we can always tighten the types later (a non-breaking change).

LauraBeatris: >  to make non-breaking changes

How it would be a non-breaking change - could you give an example? Sorry, it's not clear to me also how the current interfaces could lead to breaking changes, just trying to understand before moving forward.

LauraBeatris: Introduced a similar approach on https://github.com/workos/workos-node/pull/998/commits/39fdd375ae77605facb75fe92876b675fc9fc351

The Stripe auto-generated types include all event types as properties even as nullable. Here we're just indexing the one equivalent still as nullable to better handle breaking changes:

CleanShot 2024-03-26 at 14 52 22

LauraBeatris: Also, it seems that we're diverting a bit from the API response - it's not wrong, but just wanted to point out that this behavior will change in the future if we decide to go for auto-generated APIs, where the API spec will be almost a 1:1 map to the SDK return types. 

This is a symptom that will happen for not only authentication events but all of the other modules.

amygdalama: oh sorry, we can skip these type-specific fields for now. the API doesn't return them currently. if/when we do return them we can handle this then.

LauraBeatris:
@amygdalama Types updated! I'll proceed with testing locally.

src/user-management/interfaces/authentication-event.interface.ts
```diff
@@ -0,0 +1,30 @@
...
+  | 'sso'
+  | 'password'
+  | 'oauth'
+  | 'mfa'
+  | 'magic_auth'
+  | 'email_verification';
+
+export type AuthenticationEvent = {
+  type: AuthenticationEventType;
+  email?: string;
```
amygdalama, line 16:
```suggestion
  email: string | null;
```

src/user-management/interfaces/authentication-event.interface.ts
```diff
@@ -0,0 +1,30 @@
...
+  | 'sso'
+  | 'password'
+  | 'oauth'
+  | 'mfa'
+  | 'magic_auth'
+  | 'email_verification';
+
+export type AuthenticationEvent = {
+  type: AuthenticationEventType;
+  email?: string | null;
```
amygdalama, line 16:
```suggestion
  email: string | null;
```

src/user-management/interfaces/authentication-event.interface.ts
```diff
@@ -0,0 +1,30 @@
...
+  ipAddress: string | null;
+  userAgent: string | null;
+  userId: string | null;
+  error?: AuthenticationEventError;
+};
+
+export interface AuthenticationEventResponse {
+  type: AuthenticationEventType;
+  email?: string | null;
```
amygdalama, line 25:
```suggestion
  email: string | null;
```

LauraBeatris: Seems like we're not that consistent with nullable properties, there are a couple of other places that use `undefined | null`. 

I've updated `error` to `| null` as well.

blairlunceford:
DAAP-913

alisherry (approved):

PR #1002, opened 2024-03-28
Provide a manual workflow for fixing the npm @latest tag

jasonroelofs:
## Description
Users wanting to deploy a non-latest version will use this workflow to fix the @latest tag to the correct version once new release is published.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/.github/workflows/fix-latest.yml
+++ b/.github/workflows/fix-latest.yml
@@ -1,0 +1,36 @@
+name: Fix @latest tag
+
+on:
+  workflow_dispatch:
+    inputs:
+      latest-version:
+        description: 'The version that should be @latest, e.g. "6.7.0"'
+        required: true
+        type: string
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  test:
+    name: Update @latest tag
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 18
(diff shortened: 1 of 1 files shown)


.github/workflows/fix-latest.yml
```diff
@@ -0,0 +1,36 @@
+name: Fix @latest tag
```
amygdalama, line 1: i was wondering if there was a better way to solve this high-level problem of releasing a patch to older major versions and found this: https://stackoverflow.com/a/68297663

would there be a viable workflow for that with github actions?

either way, i think we'll want this "fix `@latest`" job around in case someone makes a mistake...

amygdalama: also, mind updating [the docs](https://www.notion.so/workos/SDK-Release-Process-a52c9c5a71ce48578e7e3e26fc82b386) when this is live?

jasonroelofs: Yeah I'm not sure there are any good choices here. We could go with major version tags, `v6`, `v5`, etc, and the release workflow can generate that automatically. However I'm not sure what that means for `latest`, I'm guessing if we only release to our own tags that `latest` won't get updated, and the workflow can't really know what the most recent is supposed to be to fix this manually.

All said, this is also a process that's relatively easy to change if we need to regularly deploy older versions and are regularly forgetting to update `latest`.

And yep, will make sure the docs are updated as well.

amygdalama (approved):

PR #1011, opened 2024-04-03
Drop support for versions < 19

jasonroelofs:
## Description
We've moved to using native `global.crypto` which only exists in newer versions of Node.js, so our tests should reflect that reality.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/setup-jest.ts
+++ b/setup-jest.ts
@@ -1,9 +1,3 @@
 import { enableFetchMocks } from 'jest-fetch-mock';
-import { Crypto } from '@peculiar/webcrypto';
 
 enableFetchMocks();
-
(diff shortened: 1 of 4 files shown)


setup-jest.ts
```diff
@@ -1,9 +1,3 @@
 import { enableFetchMocks } from 'jest-fetch-mock';
-import { Crypto } from '@peculiar/webcrypto';
 
 enableFetchMocks();
-
-// Assign Node's Crypto to global.crypto if it is not already present
```
PaulAsjes, line 6: This needs to stay in so the tests can run with Jest. See https://github.com/workos/workos-node/pull/978

PaulAsjes (changes requested):

PR #1012, opened 2024-04-03
Move the crypto polyfill into the library itself

jasonroelofs:
## Description
This lets us keep support for Node 16 and 18, while running under 19 or newer will default to the now-built-in crypto support.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,10 +34,10 @@
     "prepublishOnly": "yarn run build"
   },
   "dependencies": {
+    "@peculiar/webcrypto": "^1.4.5",
     "pluralize": "8.0.0"
   },
   "devDependencies": {
-    "@peculiar/webcrypto": "^1.4.5",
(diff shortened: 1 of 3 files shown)


hadihallak:
Would this work on Edge runtime?

jasonroelofs:
@hadihallak If I'm reading the docs right, it should be a no-op on Cloudflare Edge https://developers.cloudflare.com/workers/runtime-apis/web-crypto/

jonatascastro12:
I tested this PR on CF workers and saw it won't work smoothly. Trying to import the polyfill causes issues when importing specific Node modules. I also tried dynamic imports, but CF workers still try to import the polyfill, and the same problems happen at build time.

CleanShot 2024-04-04 at 12 40 25

Closing this PR.

PR #1013, opened 2024-04-04
Add `engines` field in `package.json` to reflect current requirements

benoitgrelard:
## Description

Add the `engines` field in `package.json` to reflect the current requirements (due to usage of Web Crypto) until we find a more long term solution to bring back backwards compatibility with older node engines.

## Documentation

```
[ ] Yes
```


package.json
```diff
@@ -12,6 +12,9 @@
     "node": "14.21.3",
     "yarn": "1.22.19"
   },
+  "engines": {
+    "node": ">=19"
```
benoitgrelard, line 16: @PaulAsjes 19 is correct right? Relating to fetch + web-crypto availability.

PaulAsjes: According to MDN, Web Crypto was fully [implemented in v19](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) and fetch was [implemented in v18](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), so requiring at least v19 makes sense to me.

PaulAsjes (approved):

benoitgrelard:
> Do we need an `"engineStrict" : true,` in there?

Looks like that could be useful?
CleanShot 2024-04-04 at 14 23 55@2x

PR #1024, opened 2024-04-12
feat: Organization events

faroceann:
## Description
Introduces support for [organization events](https://workos.com/docs/events/user-management/organization)

## Documentation

```
[x] Yes
```

### DIFF
--- a/src/organizations/interfaces/organization-domain.interface.ts
+++ b/src/organizations/interfaces/organization-domain.interface.ts
@@ -2,4 +2,7 @@ export interface OrganizationDomain {
   object: 'organization_domain';
   id: string;
   domain: string;
+  verification_token?: string;
+  verification_stragegy?: 'dns' | 'manual';
+  state?: 'failed' | 'pending' | 'legacy_verified' | 'verified';
 }
--- a/src/organizations/interfaces/organization-domain.interface.ts
+++ b/src/organizations/interfaces/organization-domain.interface.ts
@@ -2,7 +2,4 @@ export interface OrganizationDomain {
   object: 'organization_domain';
   id: string;
   domain: string;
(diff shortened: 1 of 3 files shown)


jasonroelofs (approved):

src/organizations/interfaces/organization-domain.interface.ts
```diff
@@ -2,4 +2,7 @@ export interface OrganizationDomain {
   object: 'organization_domain';
   id: string;
   domain: string;
+  verification_token?: string;
+  verification_stragegy?: 'dns' | 'manual';
```
amygdalama, line 6: typo here, but also i am wondering if we want to expose these properties in the SDK? @oliverzheng what do you think?

amygdalama: we can get into trouble when using enums / string literals since then adding a new value is a breaking API change. so perhaps we should set these to string types?

oliverzheng: Yeah we can either make it a string type, or a `'dns' | 'manual' | 'unknown'` and have `'unknown'` be a special value that deserialization uses for any value, well, unknown to the existing values. `string` is probably simple enough here though.

amygdalama: sounds like you're ok w/ exposing these 3 new properties?

rwtombaugh (approved):

src/organizations/interfaces/organization-domain.interface.ts
amygdalama: nvm i'm going to remove these for now as they aren't deserialized properly.

PR #1025, opened 2024-04-18
Add `domain_data` to Organization API and deprecate `domains`

oliverzheng:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/organizations/interfaces/update-organization-options.interface.ts
+++ b/src/organizations/interfaces/update-organization-options.interface.ts
@@ -1,12 +1,22 @@
+import { DomainData } from './domain-data.interface';
+
 export interface UpdateOrganizationOptions {
   organization: string;
   name: string;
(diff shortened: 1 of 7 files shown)


src/organizations/interfaces/update-organization-options.interface.ts
```diff
@@ -1,12 +1,22 @@
...
+
 export interface UpdateOrganizationOptions {
   organization: string;
   name: string;
   allowProfilesOutsideOrganization?: boolean;
+  /**
+   * @deprecated Use `domain_data` instead.
+   */
   domains?: string[];
+  domain_data?: DomainData[];
```
mthadley, line 11: We should use camel-case for parameters to SDK methods, and convert to snake-case as part of the serialization process.

mthadley (approved):

PR #1026, opened 2024-04-22
Update to v7.0.1

rwtombaugh:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "7.0.0",
+  "version": "7.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "7.0.0",
+  "version": "7.0.1",
```
cmatheson, line 2: i think this should be 7.1.0 since this is a feature release. (https://semver.org/)

cmatheson (changes requested):

PR #1033, opened 2024-04-25
Add explicit handling for rate-limit errors

mthadley:
## Description

Introduces a new `RateLimitExceededException` to be thrown when a `429` error is encountered, rather than defaulting to a more generic error exception class.

This also allows the reading of the `Retry-After` header, which is helpful for clients to retry, as well as debugging.

There's also some light refactoring included.

## Documentation

```
[ ] Yes
```


src/common/interfaces/request-exception.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface RequestException {
```
mthadley, line 1: In my initial pass I realized the new `RateLimitExceededException` was missing `status`. It seemed worth having something like this in order for all exception classes to be roughly shaped the same.

Arguable we should have a base-class instead, but I opted for an `interface` for now since it only adds an additional compile-time check, and we can decide later.

src/workos.ts
```diff
@@ -214,6 +215,15 @@ export class WorkOS {
...
         }
+        case 429: {
+          const retryAfter = headers.get('Retry-After');
+
+          throw new RateLimitExceededException(
+            data.message,
+            requestID,
+            retryAfter ? Number(retryAfter) : null,
+          );
+        }
```
mthadley, line 226: Is this a breaking change?

Existing users may be doing an `error instanceof GenericServerException` when attempting to handle rate limit errors, which would no longer work with the introduction of this new exception.

mthadley: Actually, one option would be to make `GenericServerException` the actual base-class that the other exceptions extend from. It can enforce that rough shape we want all of them to have, and would mean this is no longer a potential breaking-change.

Though I'd like to hear opinions before I make that change.

PaulAsjes: Strongly prefer having `GenericServerException` be the base-class for inheritance to make this not be a breaking change. We want to generally shy away from releasing major versions too frequently. 

While someone out there might be relying on the `name`, I feel that that's a bit of an anti-pattern over checking `instanceof`. I'm comfortable with releasing this as a minor version instead of a major.

src/workos.spec.ts
```diff
@@ -222,6 +224,30 @@ describe('WorkOS', () => {
       });
     });
 
+    describe('when the api responses with a 429', () => {
+      it('throws a RateLimitExceededException', async () => {
+        fetchOnce(
+          {
+            message: 'Too many requests',
```
PaulAsjes, line 231: Is this verbatim what we return from the API? I'd like there to be more information in here, like a link to the [rate limiting docs](https://workos.com/docs/reference/rate-limits) or a callout to reach out on support@ if they want to discuss higher rate limits.

src/workos.spec.ts
```diff
@@ -222,6 +224,30 @@ describe('WorkOS', () => {
...
+          },
+          {
+            status: 429,
+            headers: { 'X-Request-ID': 'a-request-id', 'Retry-After': '10' },
```
PaulAsjes, line 235: Is 10 seconds the default or is that just for this test? If it's the default returned by the API then that feels way too high.

PaulAsjes (changes requested):

src/workos.spec.ts
mthadley: It is, and agreed that we should have more information in the `message`. Separately, I can look into updating that, adding the additional details you mentioned, so that all SDK's (or no SDK at all) will have the extra context.

src/workos.spec.ts
mthadley: It's just a test value. IIRC, the actual `Retry-After` is generally larger, and varies by endpoint.

src/workos.ts
mthadley: I made the minimal change of only having the new exception inherit from `GenericServerException`, though a large change could be to have _all_ of the requests also inherit (and left a note about it): [4af4a79](https://github.com/workos/workos-node/pull/1033/commits/4af4a79e010b888466d32275650c1fb2c9faa874).

PaulAsjes (approved):

PR #1038, opened 2024-05-08
Add backwards compatibility for older versions of Node

PaulAsjes:
## Description
workos-node v6 added support for other JS runtimes, but that meant that the minimum node version required jumped up to v19.

This adds in backwards compatibility to make the SDK usable with at least Node v16. 

## Documentation

```
[ ] Yes
```


jasonroelofs:
We should also remove Node 14 from the CI matrix.

jasonroelofs:
Can probably need to tell the linter to ignore these vendored polyfills.

src/workos.spec.ts
```diff
@@ -249,14 +296,35 @@ describe('WorkOS', () => {
     });
 
     describe('when the entity is null', () => {
-      it('sends a null body', async () => {
+      it('sends an empty string body', async () => {
```
PaulAsjes, line 299: In fetch environments we can't have the body be null due to how Deno works:

https://github.com/workos/workos-node/pull/1038/files#diff-eb8ece7bf63e57536fd07d4e0ef7369670a03714e326e2a55c1c4a947993fb5dR119-R121

src/common/crypto/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './NodeCryptoProvider';
+export * from './SubtleCryptoProvider';
+export * from './CryptoProvider';
```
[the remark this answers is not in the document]
PaulAsjes: Already have! Tests are here: https://github.com/workos/workos-node/pull/1038/files#diff-6e2a0e50ff73fa9faf5240a7705aabf4e870e3c3126c877fafc136cc723d346bR227-R270

src/common/crypto/CryptoProvider.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * Interface encapsulating the various crypto computations used by the library,
+ * allowing pluggable underlying crypto implementations.
+ */
+export class CryptoProvider {
```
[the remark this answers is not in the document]
PaulAsjes: Only that I had no idea you could do that! Updated to make this and `HttpClient` abstract classes

src/webhooks/webhooks.ts
```diff
@@ -1,9 +1,22 @@
...
+  SubtleCryptoProvider,
+  NodeCryptoProvider,
+} from '../common/crypto';
 
 export class Webhooks {
-  private encoder = new TextEncoder();
+  private cryptoProvider: CryptoProvider;
+
+  constructor(subtleCrypto?: typeof crypto.subtle) {
+    if (typeof crypto.subtle !== 'undefined') {
```
alisherry, line 14: just double checking, will this be `undefined` as a string?

alisherry (approved):

src/webhooks/webhooks.ts
PaulAsjes: Yup, because `typeof` always returns the string representation of the type.

PaulAsjes:
Thanks for the review @alisherry! While doing some testing I found some bugs in the Node implementation which should now be fixed. Can you have another look?

src/common/crypto/NodeCryptoProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+      return false;
+    }
+
+    // Generate a random key for HMAC
+    const key = crypto.randomBytes(32); // Generates a 256-bit key
+    const hmacA = crypto.createHmac('sha256', key).update(bufferA).digest();
+    const hmacB = crypto.createHmac('sha256', key).update(bufferB).digest();
+
+    // Perform a constant time comparison
+    return hmacA.equals(hmacB);
```
mattgd, line 40: Is `.equals()` on a buffer actually a constant time comparison? I couldn't find any docs that confirm. Or are we able to use `crypto.timingSafeEqual()`?

PaulAsjes: Great catch, turns out GPT lied to me.

alisherry (approved):

PR #1052, opened 2024-06-03
v7.7.0

faroceann:
## Documentation

```
[ ] Yes
```


package.json
```diff
@@ -53,4 +53,4 @@
     "tslint": "6.1.3",
     "typescript": "5.1.6"
   }
-}
+}
```
amygdalama, line 56: this intentional?

amygdalama (approved):

package.json
faroceann: I think it's my auto prettier getting overzealous

PR #1053, opened 2024-06-03
Add find invitation by token method

blairlunceford:
## Description

Exposes method to find an invitation by token.

## Documentation

```
[x] Yes
```

https://github.com/workos/workos/pull/27785

### DIFF
--- a/src/user-management/user-management.spec.ts
+++ b/src/user-management/user-management.spec.ts
@@ -879,7 +879,10 @@ describe('UserManagement', () => {
       expect(fetchURL()).toContain(
         `/user_management/invitations/${invitationId}`,
       );
-      expect(invitation).toMatchObject({});
+      expect(invitation).toMatchObject({
+        object: 'invitation',
+        email: 'dane@workos.com',
+      });
     });
   });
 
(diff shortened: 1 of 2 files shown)


src/user-management/user-management.spec.ts
```diff
@@ -882,6 +883,19 @@ describe('UserManagement', () => {
...
+  describe('findInvitationByToken', () => {
+    it('sends a find invitation by token request', async () => {
+      fetchOnce(invitationFixture);
+      const invitation = await workos.userManagement.findInvitationByToken(
+        invitationToken,
+      );
+      expect(fetchURL()).toContain(
+        `/user_management/invitations/by_token/${invitationToken}`,
+      );
+      expect(invitation).toMatchObject({});
```
mattgd, line 895: Should we match a more specific deserialized object here?

blairlunceford: Good catch - I added a check that the invitation has the right token. I also updated the `getInvitation` test to check that the invitation has the right ID in that case. But we can also check against the entire object from the fixture if that's preferred.

PaulAsjes (approved):

Issue #1061, opened 2024-06-13
Could not resolve "crypto", "http", "https" in Cloudflare Pages Project


justinnoel:
# Problem
I'm unable to add the `@workos-inc/node` SDK to a newly created Cloudflare Pages project created with [C3](https://developers.cloudflare.com/pages/get-started/c3/).

When running `pnpm dev`, I get the following:

![cf-workos-errors](https://github.com/workos/workos-node/assets/81643826/04018a2e-3a06-4eed-a034-8159c90bb7df)

## Reproducible Example
See [this project](https://github.com/justinnoel/test-cf-workos) that demonstrates the issue.

## Versions
WorkOS: 7.10.0
wrangler: 3.57.2
Hono: 4.4.5
Node: 20.14.0

PaulAsjes:
This is likely due to some recent changes we made to SDK to be able to support older versions of Node. We'll investigate, but in the meanwhile if you downgrade to version 7.9.0 then it should work as expected.

justinnoel:
Thanks so much for the quick response. Downgrading to 7.9.0 did help resolve those issues.

However, now I get 

> [ERROR] ReferenceError: process is not defined

I even tried 7.8.0.  I've pushed the change to the sample repo.

![cf-workos-errors-process](https://github.com/workos/workos-node/assets/81643826/38fe910c-2f4c-45dd-81f2-fa0d718e986b)

reecejohnson:
I was also experiencing this same issue on version `7.8.0`

Upgrading to `7.9.0` fixed it for me too. 

Thanks!

danjenkins:
@justinnoel I'm currently getting process is not defined... what did you to fix it? Do you have to set node_compat in cloudflare workers?

justinnoel:
@danjenkins All I needed to do was install `"@workos-inc/node": "7.9.0"`.

I didn't use `node_compat`.

danjenkins:
How strange - I'm getting process is undefined in 7.14 and tried 7.9... assume it worked fine in wrangler dev and you didn't have to deploy it?

justinnoel:
@danjenkins No problems in wrangler for 7.9.  Once I solved that in dev, I deployed to preview and production and it worked fine.

danjenkins:
Weird! I've made a new issue... will see what the workos team say - thanks for replying!

danjenkins:
Wondering if your example being in typescript has anything to do with anything... 

Mine is just pure JavaScript

PR #1062, opened 2024-06-17
Fixed cloudflare worker support

PaulAsjes:
## Description
Fixes https://github.com/workos/workos-node/issues/1061

Cloudflare Workers were failing because CFWs don't allow dynamic requiring of Node specific libraries. This adopts a new strategy where we have a CFW specific entry (`index.worker.js`) that doesn't try to select a Crypto/HTTP library at runtime.

## Documentation

```
[ ] Yes
```


package.json
```diff
@@ -51,5 +50,14 @@
...
     "typescript": "5.1.6"
+  },
+  "exports": {
+    "types": "./types/index.d.ts",
+    "worker": {
+      "import": "./lib/index.worker.js"
+    },
+    "default": {
+      "import": "./lib/index.js"
+    }
```
mattgd, line 61: I've never used the exports key before. How does this work? Does developers need to import it a certain way to use the worker variant or is it selected automagically?

PaulAsjes: It works automagically, which I was pretty surprised about. It basically does some under the hood runtime detection and picks the entry point that you've defined. I tested this in node 16 & 20, bun, deno and CFWs and it all worked as expected. Pretty neat.

mattgd (approved):

package.json
Leestex: @mattgd @PaulAsjes please take a look at [this PR](https://github.com/workos/workos-node/pull/1068), as this commit caused breaking changes

PR #1071, opened 2024-07-02
Fix `NodeHttpClient` handling of `URLSearchParam` request bodies 

mthadley:
## Description

Previously the client was passing the body directly down to the low level `write` and `byteLength` calls, which failed in cases where the body wasn't already serialized as a `string` or form of buffer.

This was not noticed on newer versions of Node since they include a global `fetch`, resulting in the SDK preferring the `FetchHttpClient`.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/common/net/node-client.ts
+++ b/src/common/net/node-client.ts
@@ -40,6 +40,18 @@ export class NodeHttpClient extends HttpClient implements HttpClientInterface {
     return 'node';
   }
 
+  static override getBody(entity: unknown): string | null {
+    if (entity === null || entity === undefined) {
+      return null;
+    }
+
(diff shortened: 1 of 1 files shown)


src/common/net/node-client.ts
```diff
@@ -40,6 +40,18 @@ export class NodeHttpClient extends HttpClient implements HttpClientInterface {
     return 'node';
   }
 
+  static override getBody(entity: unknown): string | null {
```
mthadley, line 43: I decided to override the parent class's `getBody` since it's return value of `BodyInit` (which is compatible with `fetch`) didn't make sense for the Node HTTP API.

cmatheson (approved):

PR #1077, opened 2024-07-10
Add module to support FGA endpoints

stanleyphu:
## Description
This PR adds a new FGA module with support for FGA endpoints, including `objects`, `warrants`, `check`, and `query`.

## Documentation

```
[x] Yes
```

https://github.com/workos/workos/pull/28797


src/workos.ts
```diff
@@ -35,6 +36,10 @@ const VERSION = '7.14.0';
 
 const DEFAULT_HOSTNAME = 'api.workos.com';
 
+const AUTHORIZATION_HEADER = 'Authorization';
```
kkajla12, line 39: Nitpick, but my personal preference on naming constant like this is:
```javascript
const HEADER_AUTHORIZATION = 'Authorization';
const HEADER_IDEMPOTENCY_KEY = 'Idempotency-Key';
const HEADER_WARRANT_TOKEN = 'Warrant-Token';
```

kkajla12 (approved):

kkajla12 (approved):

src/fga/fga-live-test.spec.ts
```diff
@@ -0,0 +1,1136 @@
+import { disableFetchMocks, enableFetchMocks } from 'jest-fetch-mock';
+
+import { WorkOS } from '../workos';
+import { CheckOp, WarrantOp } from './interfaces';
+
+describe.skip('FGA Live Test', () => {
```
mthadley, line 6: Is this `.skip` intentional? I'm guessing yes, given the `live` naming, but just wanted to double-check.

src/fga/interfaces/query.interface.ts
```diff
@@ -0,0 +1,33 @@
...
+  context?: string;
+}
+
+export interface QueryResult {
+  resourceType: string;
+  resourceId: string;
+  relation: string;
+  warrant: Warrant;
+  isImplicit: boolean;
+  meta?: { [key: string]: any };
```
mthadley, line 21: Nit: We'll typically reach for `Record` instead of declaring an object with an index signature:

```suggestion
  meta?: Record<string, any>;
```

src/fga/serializers/check-options.serializer.ts
```diff
@@ -0,0 +1,76 @@
...
+    subject: isSubject(warrant.subject)
+      ? {
+          resource_type: warrant.subject.resourceType,
+          resource_id: warrant.subject.resourceId,
+        }
+      : {
+          resource_type: warrant.subject.getResourceType(),
+          resource_id: warrant.subject.getResourceId(),
+        },
+    context: warrant.context ? warrant.context : {},
```
mthadley, line 51: Nit:

```suggestion
    context: warrant.context ?? {},
```

src/fga/utils/interface-check.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export function isResourceInterface(
+  resource: any,
+): resource is ResourceInterface {
+  return (
+    resource.getResourceType !== undefined &&
+    resource.getResourceId !== undefined
+  );
+}
```
mthadley, line 17: We typically prefer `unknown` over `any`. In this case, the `any` would allow `null` or `undefined` bass which would cause this function to throw.

How about this?

```suggestion
export function isResourceInterface(
  resource: unknown,
): resource is ResourceInterface {
  return (
    !!resource &&
    typeof resource === 'object' &&
    'getResourceType' in resource &&
    'getResourceId' in resource
  );
}
```

src/workos.ts
```diff
@@ -141,13 +151,20 @@ export class WorkOS {
     path: string,
     options: GetOptions = {},
   ): Promise<{ data: Result }> {
+    const requestHeaders: any = {};
```
mthadley, line 154: I know this was just copied from the other helpers, so don't feel the need to change this, but I'd love if we replaced these `any` `requestHeaders` with `Record<string, string>` instead:

```suggestion
    const requestHeaders: Record<string, string> = {};
```

mthadley (approved):
Left a few comments but all really just nitpicks. Overall, LGTM!

src/fga/fga-live-test.spec.ts
stanleyphu: Correct. Ideally we'll run this in CI in the future but right now we're only running it manually

PR #1091, opened 2024-07-29
Fix iron-session worker support

mattgd:
## Description
Fix iron-session worker support.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/setup-jest.ts
+++ b/setup-jest.ts
@@ -1,9 +1,16 @@
 import { enableFetchMocks } from 'jest-fetch-mock';
 import { Crypto } from '@peculiar/webcrypto';
+import { WorkOS } from './src/workos';
+import { WebIronSessionProvider } from './src/common/iron-session/web-iron-session-provider';
 
 enableFetchMocks();
 
 // Assign Node's Crypto to global.crypto if it is not already present
 if (!global.crypto) {
-    global.crypto = new Crypto();
+  global.crypto = new Crypto();
 }
+
+// For tests, we can use the WebIronSessionProvider
+WorkOS.prototype.createIronSessionProvider = jest
+  .fn()
+  .mockReturnValue(new WebIronSessionProvider());
--- a/src/common/iron-session/iron-session-provider.ts
+++ b/src/common/iron-session/iron-session-provider.ts
@@ -11,6 +11,11 @@ export type UnsealedDataType = Record<string, unknown>;
 
 /**
  * Interface encapsulating the sealData/unsealData methods for separate iron-session implementations.
+ *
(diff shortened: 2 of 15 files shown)


src/common/iron-session/web-iron-session-provider.ts
```diff
@@ -0,0 +1,25 @@
...
+} from './iron-session-provider';
+
+/**
+ * WebIronSessionProvider which uses the base iron-session seal/unseal methods.
+ */
+export class WebIronSessionProvider extends IronSessionProvider {
+  /** @override */
+  async sealData(data: unknown, options: SealDataOptions): Promise<string> {
+    const ironSession = await import('iron-session');
+    return ironSession.sealData(data, options);
```
mattgd, line 14: Using dynamic imports here because the problematic crypto import is in `iron-session`'s index.js.

src/workos.ts
```diff
@@ -114,6 +122,18 @@ export class WorkOS {
     }) as HttpClient;
   }
 
+  createIronSessionProvider(): IronSessionProvider {
```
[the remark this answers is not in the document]
mattgd: Updated!

setup-jest.ts
```diff
@@ -1,9 +1,16 @@
...
 // Assign Node's Crypto to global.crypto if it is not already present
 if (!global.crypto) {
-    global.crypto = new Crypto();
+  global.crypto = new Crypto();
 }
+
+// For tests, we can use the WebIronSessionProvider
+WorkOS.prototype.createIronSessionProvider = jest
+  .fn()
+  .mockReturnValue(new WebIronSessionProvider());
```
mattgd, line 16: Don't love this testing edit. Alternatively I could pass the iron-session-provider in the constructor WorkOS, but feels a bit odd given we only use it for User Management.

PaulAsjes: I think it's fine, it would be weird if we required it when you're only interested in SSO or one of the other products.

src/common/iron-session/iron-session-provider.ts
```diff
@@ -0,0 +1,22 @@
...
+    | {
+        [id: string]: string;
+      };
+  ttl?: number | undefined;
+};
+
+export type UnsealedDataType = Record<string, unknown>;
+
+/**
+ * Interface encapsulating the sealData/unsealData methods for separate iron-session implementations.
```
PaulAsjes, line 13: Let's add a comment for future generations on why we're doing this, plus that we can effectively deprecate this when we drop support for Node 16.

mattgd: Adding!

PR #1092, opened 2024-07-29
Add test for worker environment

PaulAsjes:
## Description
Adds a test in Miniflare, Cloudflare's dev environment for workers. Idea is that if the test runs and passes then the SDK should work in woeker environments.

## Documentation

```
[ ] Yes
```


mattgd:
Should the test be failing now with CFW support broken in main?

mattgd (approved):
Thanks. We can add test file in follow up.

Issue #1081, opened 2024-07-22
Cloudflare Worker - ReferenceError: process is not defined


danjenkins:
Using `@workos/node` v7.14 in a cloudflare worker locally using `wrangler dev` and being given an error.

using node 22
wrangler 3.65.1

```
ReferenceError: process is not defined
at new WorkOS2
  (file:///Users/danjenkins/<path>/node_modules/@workos-inc/node/lib/workos.js:47:24)
      at new WorkOSWorker
  (file:///Users/danjenkins/<path>/node_modules/@workos-inc/node/lib/index.worker.js:34:1)
      at Array.<anonymous>
```

Am I meant to use compatibility modes in cloudflare workers?

danjenkins:
Worked it out, I was passing undefined to the WorkOS constructor... so it would end up getting into a bit of codebase where process was indeed not defined. Feels like the module shouldn't get to that point though!

PaulAsjes:
My guess is you ran afoul of this line here: https://github.com/workos/workos-node/blob/main/src/workos.ts/#L57

We should add a check on whether `process` is defined there and bail if it isn't. I'll leave this issue open as a reminder to fix that. Thanks!

mattgd:
This should be fixed start in [v7.20.0](https://github.com/workos/workos-node/releases/tag/v7.20.0). Closing this issue out, but feel free to open another issue if you're still seeing errors. Thanks again for the report!

PR #1096, opened 2024-07-30
Use typeof to safely check Node process on workers

mattgd:
## Description
Use typeof to safely check Node process on workers.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/workos.ts
+++ b/src/workos.ts
@@ -62,7 +62,10 @@ export class WorkOS {
   constructor(readonly key?: string, readonly options: WorkOSOptions = {}) {
     if (!key) {
       // process might be undefined in some environments
-      this.key = process?.env.WORKOS_API_KEY;
+      this.key =
+        typeof process !== 'undefined'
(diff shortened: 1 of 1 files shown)


src/workos.ts
```diff
@@ -73,7 +76,10 @@ export class WorkOS {
       this.options.https = true;
     }
 
-    this.clientId = this.options.clientId ?? process?.env.WORKOS_CLIENT_ID;
+    this.clientId = this.options.clientId;
+    if (!this.clientId && typeof process !== 'undefined') {
+      this.clientId = process?.env.WORKOS_CLIENT_ID;
+    }
```
mattgd, line 82: Looked in to how we could replicate the lack of `process` in a jest test to no avail. May be something we want to dig into further to strengthen the worker test.

gcarvelli (approved):

PaulAsjes (approved):
Neat, this'll also fix https://github.com/workos/workos-node/issues/1081

PR #1100, opened 2024-08-06
Update `refreshAndSealSessionData` to accept `organizationId`

mthadley:
## Description

This matches the `authenticateWithRefreshToken` method which also accepts an `organizationId` to allow "switching".

## Documentation

```
[ ] Yes
```


src/user-management/interfaces/refresh-and-seal-session-data.interface.ts
```diff
@@ -5,8 +5,14 @@ export enum RefreshAndSealSessionDataFailureReason {
...
   NO_SESSION_COOKIE_PROVIDED = 'no_session_cookie_provided',
+
   // API OauthErrors for refresh tokens
   INVALID_GRANT = 'invalid_grant',
+  MFA_ENROLLMENT = 'mfa_enrollment',
+  SSO_REQUIRED = 'sso_required',
+  /**
+   * @deprecated To be removed in a future major version.
+   */
   ORGANIZATION_NOT_AUTHORIZED = 'organization_not_authorized',
```
mthadley, line 16: This is not actually an error that can be returned.

PaulAsjes (approved):

PR #1105, opened 2024-08-16
Rework session helpers

PaulAsjes:
## Description

Refactors the session helpers to add a new session class. The old helper methods will stay put, but we'll likely add a deprecation comment and remove them in a future major version.

See this [gist](https://gist.github.com/PaulAsjes/e1f80d5451f988db2acdf6ee894fcc46) as an example of how this would work in practise. 

## Documentation

```
[x] Yes
```

### DIFF
--- a/src/user-management/interfaces/authenticate-with-session-cookie.interface.ts
+++ b/src/user-management/interfaces/authenticate-with-session-cookie.interface.ts
@@ -36,6 +36,6 @@ export type AuthenticateWithSessionCookieSuccessResponse = {
   organizationId?: string;
   role?: string;
   permissions?: string[];
-  user?: User;
+  user: User;
   impersonator?: Impersonator;
 };
--- a/src/user-management/interfaces/authenticate-with-session-cookie.interface.ts
+++ b/src/user-management/interfaces/authenticate-with-session-cookie.interface.ts
@@ -1,4 +1,6 @@
 import { AuthenticationResponse } from './authentication-response.interface';
+import { Impersonator } from './impersonator.interface';
+import { User } from './user.interface';
 
 export interface AuthenticateWithSessionCookieOptions {
   sessionData: string;
@@ -34,4 +36,6 @@ export type AuthenticateWithSessionCookieSuccessResponse = {
   organizationId?: string;
   role?: string;
   permissions?: string[];
+  user?: User;
+  impersonator?: Impersonator;
 };
--- a/src/user-management/interfaces/refresh-and-seal-session-data.interface.ts
+++ b/src/user-management/interfaces/refresh-and-seal-session-data.interface.ts
@@ -18,6 +18,7 @@ export enum RefreshAndSealSessionDataFailureReason {
   ORGANIZATION_NOT_AUTHORIZED = 'organization_not_authorized',
 }
 
+// TODO: These should be renamed since it's possible to have an unsealed session
 type RefreshAndSealSessionDataFailedResponse = {
   authenticated: false;
   reason: RefreshAndSealSessionDataFailureReason;
--- a/src/user-management/interfaces/refresh-and-seal-session-data.interface.ts
+++ b/src/user-management/interfaces/refresh-and-seal-session-data.interface.ts
@@ -1,3 +1,5 @@
+import { AuthenticationResponse } from './authentication-response.interface';
+
 export enum RefreshAndSealSessionDataFailureReason {
   /**
    * @deprecated To be removed in a future major version.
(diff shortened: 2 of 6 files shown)


src/user-management/interfaces/authenticate-with-session-cookie.interface.ts
```diff
@@ -34,4 +36,6 @@ export type AuthenticateWithSessionCookieSuccessResponse = {
   organizationId?: string;
   role?: string;
   permissions?: string[];
+  user?: User;
```
mattgd, line 39: Is user guaranteed on successful auth, or no?

src/user-management/user-management.ts
```diff
@@ -511,7 +524,7 @@ export class UserManagement {
     authenticationResponse: AuthenticationResponse;
     session?: AuthenticateWithSessionOptions;
   }): Promise<AuthenticationResponse> {
-    if (session) {
+    if (session?.sealSession) {
```
mattgd, line 527: 🙌

src/user-management/user-management.ts
```diff
@@ -155,19 +156,28 @@ const toQueryString = (options: Record<string, string | undefined>): string => {
...
+
     // Set the JWKS URL. This is used to verify if the JWT is still valid
     this.jwks = clientId
       ? createRemoteJWKSet(new URL(this.getJwksUrl(clientId)))
       : undefined;
   }
 
+  session(sessionData: string, cookiePassword: string): Session {
+    return new Session(this, sessionData, cookiePassword);
+  }
```
mattgd, line 179: I wonder if this should be called something like `sessionFromCookie()`. Also do we want to allow `WORKOS_COOKIE_PASSWORD` from environment variable at all? We currently do this, but I don't have a strong opinion either way.

src/user-management/user-management.ts
```diff
@@ -484,7 +494,10 @@ export class UserManagement {
         };
       }
 
-      return { authenticated: true, sealedSession };
+      return {
+        authenticated: true,
+        session: sealedSession,
+      };
```
mattgd, line 500: Breaking change?

src/user-management/session.ts
```diff
@@ -0,0 +1,202 @@
...
+          organizationId:
+            options.organizationId ?? organizationIdFromAccessToken,
+          session: {
+            // We want to store the new sealed session in this class instance, so this always needs to be true
+            sealSession: true,
+            cookiePassword: options.cookiePassword ?? this.cookiePassword,
+          },
+        });
+
+      if (options.sealSession) this.cookiePassword = options.cookiePassword;
```
mattgd, line 149: I'm strongly pro-curly braces for maintainability.
https://www.imperialviolet.org/2014/02/22/applebug.html
```suggestion
      if (options.sealSession) {
        this.cookiePassword = options.cookiePassword;
      }
```

src/user-management/interfaces/refresh-and-seal-session-data.interface.ts
```diff
@@ -16,14 +18,15 @@ export enum RefreshAndSealSessionDataFailureReason {
...
+// TODO: These should be renamed since it's possible to have an unsealed session
 type RefreshAndSealSessionDataFailedResponse = {
   authenticated: false;
   reason: RefreshAndSealSessionDataFailureReason;
 };
 
 type RefreshAndSealSessionDataSuccessResponse = {
   authenticated: true;
-  sealedSession: string;
+  session: string | AuthenticationResponse;
```
mattgd, line 29: Breaking change?

src/user-management/user-management.ts
PaulAsjes: I kinda like the terseness of `session()`.

I want to shy away from using environment variables as it causes headaches in environments that don't support `process` (looking at you CFW).

src/user-management/interfaces/refresh-and-seal-session-data.interface.ts
PaulAsjes: Bah, you're right. I think we'll need a `sealedSession` and `session` to not break this.

src/user-management/user-management.ts
mattgd: Good point on CFW, works for me!

src/user-management/interfaces/authenticate-with-session-cookie.interface.ts
PaulAsjes: Should do, can't think of a scenario with successfully authenticating doesn't get you the user object.

mattgd (approved):
Looks good, thanks for updating! Left one follow up comment.

PR #1108, opened 2024-08-19
fix: Ensure organizationId is included in SessionCookieData during sealSessionDataFromAuthenticationResponse

cmanou:
## Description
They types were a slight lie, since you could have had a session with an org id in the access token the but the organization id in the sealed cookie data didn't include it. This required users to always decode the jwt even in cases where they were trying to abstract away using the new sealed session data. 

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/user-management/user-management.spec.ts
+++ b/src/user-management/user-management.spec.ts
@@ -142,7 +142,11 @@ describe('UserManagement', () => {
 
     describe('when sealSession = true', () => {
       beforeEach(() => {
-        fetchOnce({ user: userFixture });
+        fetchOnce({
+          user: userFixture,
+          access_token:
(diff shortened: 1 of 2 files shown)


PaulAsjes:
Thanks for this! Looks like the tests failed though, would you mind looking at those before I approve this?

cmanou:
@PaulAsjes apologies, had some issues getting the tests running with node versions, the rogue nvmrc with 14 gave me some grief aswell as working out which pm to use. But should be all good now. I just updated the fixtures to use the same access token that was already in the file let me know if you want to change it any way. I opted not to change the teat outputs to much given  they weren't previously testing the access tokens and the upcoming session helper rework

PaulAsjes:
No worries, I think we're almost there. Just need you to run `npm run format` and push the changes to fix the prettier issues and I think we should be good.

cmanou:
Should be all set now

PaulAsjes (approved):

PR #1114, opened 2024-08-27
Add 'scrypt' to passwordHashType in user options interfaces

jonatascastro12:


src/user-management/interfaces/create-user-options.interface.ts
```diff
@@ -2,7 +2,7 @@ export interface CreateUserOptions {
   email: string;
   password?: string;
   passwordHash?: string;
-  passwordHashType?: 'bcrypt' | 'firebase-scrypt' | 'ssha';
+  passwordHashType?: 'bcrypt' | 'firebase-scrypt' | 'ssha' | 'scrypt';
```
mattgd, line 5: Should abstract these hash types into its own type since it's used in several spots?

mattgd (approved):

src/user-management/interfaces/create-user-options.interface.ts
jonatascastro12: Sounds good.

PR #1134, opened 2024-09-30
Add network retries for FGA endpoints 

stanleyphu:
## Description

- Adds network retries for FGA endpoints to `FetchHttpClient`

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/common/net/fetch-client.ts
+++ b/src/common/net/fetch-client.ts
@@ -7,6 +7,13 @@ import {
 } from '../interfaces/http-client.interface';
 import { HttpClient, HttpClientError, HttpClientResponse } from './http-client';
 
+const MAX_RETRY_ATTEMPTS = 3;
+const BACKOFF_MULTIPLIER = 1.5;
+const MINIMUM_SLEEP_TIME = 500;
+const RETRY_STATUS_CODES = [500, 502, 504];
+
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
 export class FetchHttpClient extends HttpClient implements HttpClientInterface {
   private readonly _fetchFn;
 
@@ -45,7 +52,16 @@ export class FetchHttpClient extends HttpClient implements HttpClientInterface {
       options.params,
     );
 
-    return await this.fetchRequest(resourceURL, 'GET', null, options.headers);
+    if (resourceURL.includes('fga/')) {
+      return await this.fetchRequestWithRetry(
+        resourceURL,
+        'GET',
+        null,
+        options.headers,
+      );
+    } else {
+      return await this.fetchRequest(resourceURL, 'GET', null, options.headers);
+    }
   }
 
   async post<Entity = any>(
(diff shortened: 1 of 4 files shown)


src/common/net/fetch-client.ts
```diff
@@ -151,6 +200,93 @@ export class FetchHttpClient extends HttpClient implements HttpClientInterface {
 
     return new FetchHttpClientResponse(res);
   }
+
+  private async fetchRequestWithRetry(
```
kkajla12, line 204: @stanleyphu, is there any way we could wrap the `fetchRequest` method above to avoid having to duplicate most of its logic? For example, I wonder if the section of code inside the `try` block could call `fetchRequest`

src/common/net/fetch-client.ts
```diff
@@ -151,13 +200,74 @@ export class FetchHttpClient extends HttpClient implements HttpClientInterface {
...
+    url: string,
+    method: string,
+    body?: any,
+    headers?: RequestHeaders,
+  ): Promise<HttpClientResponseInterface> {
+    let response: HttpClientResponseInterface;
+    let requestError: any = null;
```
kkajla12, line 211: We should declare `requestError` inside of `makeRequest`. It doesn't need to be outside the scope of that method.

src/common/net/fetch-client.ts
```diff
@@ -151,13 +200,74 @@ export class FetchHttpClient extends HttpClient implements HttpClientInterface {
...
+    let retryAttempts = 1;
+
+    const makeRequest = async (): Promise<HttpClientResponseInterface> => {
+      try {
+        response = await this.fetchRequest(url, method, body, headers);
```
kkajla12, line 216: Nice! Now we only need to maintain the one version of `fetchRequest`. Can we do the same thing for the node client as well?

kkajla12: Discussed via Slack that doing this for the node client would require a bigger refactor of the `nodeRequest` method (since the node client uses callbacks instead of async/await). To de-risk this PR (already quite large/risky), we're not going to do this.

src/common/net/fetch-client.ts
```diff
@@ -45,7 +45,16 @@ export class FetchHttpClient extends HttpClient implements HttpClientInterface {
       options.params,
     );
 
-    return await this.fetchRequest(resourceURL, 'GET', null, options.headers);
+    if (resourceURL.includes('fga/')) {
```
mattgd, line 48: Not a huge deal since if the retries work we could use it for other requests, but is there a safer way we can check for `fga` in the URL using some sort of `startsWith` logic? I worry a resource ID could include `fga` and take this path.

src/common/net/fetch-client.ts
```diff
@@ -151,6 +193,61 @@ export class FetchHttpClient extends HttpClient implements HttpClientInterface {
...
+
+  private shouldRetryRequest(requestError: any, retryAttempt: number): boolean {
+    if (retryAttempt > this.MAX_RETRY_ATTEMPTS) {
+      return false;
+    }
+
+    if (requestError != null) {
+      if (requestError instanceof TypeError) {
+        return true;
+      }
```
mattgd, line 239: What's an example of a TypeError we'd get here that we'd want to retry?

src/common/net/http-client.ts
```diff
@@ -7,6 +7,11 @@ import {
 } from '../interfaces/http-client.interface';
 
 export abstract class HttpClient implements HttpClientInterface {
+  readonly MAX_RETRY_ATTEMPTS = 3;
+  readonly BACKOFF_MULTIPLIER = 1.5;
+  readonly MINIMUM_SLEEP_TIME = 500;
```
mattgd, line 12: Nit: might be nice to denote as milliseconds.
```suggestion
  readonly MINIMUM_SLEEP_TIME_IN_MILLISECONDS = 500;
```

src/common/net/http-client.ts
```diff
@@ -82,6 +87,18 @@ export abstract class HttpClient implements HttpClientInterface {
 
     return JSON.stringify(entity);
   }
+
+  private getSleepTime(retryAttempt: number): number {
```
mattgd, line 91: Same nit re: denoting time scale
```suggestion
  private getSleepTimeInMilliseconds(retryAttempt: number): number {
```

src/common/net/fetch-client.ts
stanleyphu: The main errors we're trying to handle here are request timeouts that happen if a check or query take too long. One that came up recently was: `Unexpected error: TypeError: fetch failed` which happened because a 499 was returned after 5s.

Here's the full list of errors `fetch` throws a `TypeError` for: https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#exceptions

mattgd (approved):

PR #1145, opened 2024-10-24
Fix organization domain interface

amygdalama:
## Description
* Makes `verificationToken` optional (it was always optional from the API's perspective)
* Correctly deserializes `verificationToken` and `verificationStrategy` when deserializing organizations

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/organizations/serializers/organization.serializer.ts
+++ b/src/organizations/serializers/organization.serializer.ts
@@ -1,3 +1,4 @@
+import { deserializeOrganizationDomain } from '../../organization-domains/serializers/organization-domain.serializer';
 import { Organization, OrganizationResponse } from '../interfaces';
 
 export const deserializeOrganization = (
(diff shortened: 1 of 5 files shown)


src/organizations/serializers/organization.serializer.ts
```diff
@@ -8,7 +9,7 @@ export const deserializeOrganization = (
   name: organization.name,
   allowProfilesOutsideOrganization:
     organization.allow_profiles_outside_organization,
-  domains: organization.domains,
+  domains: organization.domains.map(deserializeOrganizationDomain),
```
amygdalama, line 12: previously the domains were not deserialized and as a result had fields like `verification_token` and `verification_strategy` instead of `verificationToken` and `verificationStrategy`. this is why i updated the test above.

PaulAsjes (approved):

Issue #1149, opened 2024-10-31
TypeError: fetch failed


oskarwilliams:
# Overview

We have seen a few cases in the past month where our calls to list events has resulted in an `Unexpected error: TypeError: fetch failed `. It would be very useful if you could add the cause to this error so that it would propagate through and allow us to see what the error is, and whether it is solvable on our end.

# Error
```
{
    message: Unexpected error: TypeError: fetch failed,
    name: Error,
    stack:
Error: Unexpected error: TypeError: fetch failed at WorkOSNode.handleHttpError (/app/node_modules/@workos-inc/node/lib/workos.js:179:19) at WorkOSNode.<anonymous> (/app/node_modules/@workos-inc/node/lib/workos.js:135:22) at Generator.throw (<anonymous>) at rejected (/app/node_modules/@workos-inc/node/lib/workos.js:6:65) at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
}
```

# Cause/Solution

The error is being thrown here https://github.com/workos/workos-node/blob/main/src/workos.ts#L236, where instead it could be ``new Error(`Unexpected error: ${error}`, { cause: error })``

PaulAsjes:
Hey there, that's a reasonable request. I'll look into this soon.

PR #1152, opened 2024-11-04
Update TS target and provide better error messagins for http errors

PaulAsjes:
## Description
Fixes https://github.com/workos/workos-node/issues/1149

Had to target a more recent version of ES to make TypeScript recognize the `options` argument of the [Error constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error).

## Documentation

```
[ ] Yes
```

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,7 +14,7 @@
     "declaration": true,
     "outDir": "lib",
     "strict": true,
-    "lib": ["dom", "es2019"],
+    "lib": ["dom", "es2022"],
     "types": ["jest", "jest-environment-miniflare/globals"]
   },
   "include": ["src"]
--- a/src/workos.ts
+++ b/src/workos.ts
@@ -233,7 +233,7 @@ export class WorkOS {
 
   private handleHttpError({ path, error }: { path: string; error: unknown }) {
     if (!(error instanceof HttpClientError)) {
-      throw new Error(`Unexpected error: ${error}`);
+      throw new Error(`Unexpected error: ${error}`, { cause: error });
     }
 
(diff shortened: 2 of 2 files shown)


tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "declaration": true,
     "outDir": "lib",
     "strict": true,
-    "lib": ["dom", "es2019"],
+    "lib": ["dom", "es2022"],
```
gcarvelli, line 17: Does this change break compatibility for users targeting older JS versions?

PaulAsjes: Claude tells me it's okay since this only affects type checking:

> The "lib" option in tsconfig.json specifies which built-in TypeScript type definitions your project can use. It determines which JavaScript APIs and features are available to your code. In your case:
> 
> * "dom" - Provides type definitions for browser APIs (like document, window, HTMLElement, etc.)
> * "es2022" - Provides type definitions for ECMAScript 2022 features (including Error cause, Array.at(), Object.hasOwn(), etc.)
> 
> Without specifying these libraries:
> 
> 1. TypeScript wouldn't know about browser-specific APIs (from "dom")
> 2. TypeScript wouldn't recognize newer JavaScript features (from "es2022")
> 
> For example, if you only had "es2019" instead of "es2022", the Error constructor with cause would give you a type error because that feature wasn't available until ES2022.
> 
> It's important to note that "lib" only affects TypeScript's type checking - it doesn't affect how your code is compiled. The actual JavaScript features available at runtime are determined by your "target" setting and your runtime environment.

gcarvelli (approved):

Issue #1130, opened 2024-09-17
Polyfill `buffer` usage breaks Cloudflare Workers `nodejs_compat_v2`


cayter:
Cloudflare Workers will now support better NodeJS compatibility as announced [here](https://blog.cloudflare.com/more-npm-packages-on-cloudflare-workers-combining-polyfills-and-native-code/).

However, the package currently relies on `iron-webcrypto` that uses `buffer` [polyfill](https://www.npmjs.com/package/buffer) and it leads to `nodejs_compat_v2` facing the issue below:

<img width="919" alt="Screenshot 2024-09-17 at 12 47 11 PM" src="https://github.com/user-attachments/assets/ef553580-acae-4b1b-adb1-584c92cfbb35">

EnriqueRuvalcaba:
# Issue: `buffer` Module Path Error When Using Cloudflare Workers with Node Compatibility v2
## Overview

I'm encountering an issue when using the `@workos-inc/node` package in a project running on Cloudflare Workers with `nodejs_compat_v2`. The problem occurs due to a path-related error originating from the `buffer` module when trying to deploy or run the worker in a development environment.

The error specifically arises when the `compatibility_date` in `wrangler.toml` is set to September 23rd, 2024, or later, which activates `nodejs_compat_v2`. This issue doesn't occur with the older `nodejs_compat` version, but I cannot use the older compatibility mode due to other dependencies in my actual project that require the latest version.

## Tech Stack

- **Node.js**: v18+
- **Cloudflare Workers**
- **Hono.js**: v4.6.3
- **Wrangler**: v3.78.11
- **@workos-inc/node**: v7.27.3

## Project Description

This is a simple Cloudflare Worker that utilizes the `@workos-inc/node` library to retrieve a list of organizations via an API call. The Worker is set up using `Hono.js`, and the environment variables are injected using Cloudflare's bindings.

Below is the code that reproduces the error:

### `package.json`

```json
{
  "name": "workos-error",
  "scripts": {
    "dev": "wrangler dev src/index.ts",
    "deploy": "wrangler deploy --minify src/index.ts"
  },
  "dependencies": {
    "@workos-inc/node": "^7.27.3",
    "hono": "^4.6.3"
  },
  "devDependencies": {
    "@cloudflare/workers-types": "^4.20240529.0",
    "wrangler": "^3.57.2"
  }
}
```

### `src/index.ts`

```typescript
import { WorkOS } from "@workos-inc/node";
[...]

PR #1156, opened 2024-11-06
Testing new iron-session package version

fforres:
## Description

Related to https://github.com/workos/workos-node/issues/1130

Passes all tests, however finding errors on building the final package (using current projects typescript version). As iron-session built package specifies one of their types to be a generic. 

```
node_modules/.pnpm/iron-session@8.0.3/node_modules/iron-session/dist/index.d.cts:112:77 - error TS2315: Type 'ServerResponse' is not generic.

112     <T extends object>(req: http.IncomingMessage | Request, res: Response | http.ServerResponse<http.IncomingMessage>, sessionOptions: SessionOptions): Promise<IronSession<T>>;
                                                                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I did manually remove the offending type from the created file, built/packaged/released on my account: https://www.npmjs.com/package/@fforres/workos-inc-node and from my tests on a monorepo, seems to be working.

If this helps moving

## Documentation

```
[ ] Yes
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "prepublishOnly": "yarn run build"
   },
   "dependencies": {
-    "iron-session": "~6.3.1",
+    "@types/cookie": "^0.6.0",
+    "iron-session": "~8.0.3",
     "jose": "~5.6.3",
     "pluralize": "8.0.0"
   },
--- a/src/common/iron-session/edge-iron-session-provider.ts
+++ b/src/common/iron-session/edge-iron-session-provider.ts
@@ -1,4 +1,4 @@
-import { sealData, unsealData } from 'iron-session/edge';
+import { sealData, unsealData } from 'iron-session';
 import {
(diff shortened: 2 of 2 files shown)


matheins:
@stanleyphu this would really unblock the cloudflare worker community

fforres:
Been testing it for a couple of weeks now in a real application and no issues. Real fix should be different, but would help the many-many folks using workers 🙏🏼 

Also pinging  @mattgd @gcarvelli JIC?

PaulAsjes:
Thanks for the PR @fforres! Problem at the moment is that we need to support Node v16, which iron-session v8 does not if memory serves. We try to circumvent this by providing a [worker specific entry point](https://github.com/workos/workos-node/blob/main/src/index.worker.ts), but clearly something isn't working as expected.

PR #1161, opened 2024-11-08
feat: Add actions support

faroceann:
## Description
Introduce new methods for Actions:
- `signResponse`: Supports forming and signing actions responses from an actions endpoint
- `verifyHeader`: Supports incoming requests from WorkOS reaching actions endpoints  

Example usage:
```typescript
app.post("/registration-action", async (req, res) => {
  const context = req.body;

  // Verify incoming workos-signature 
  try {
    await workos.actions.verifyHeader({
      payload: context,
      sigHeader: req.headers["workos-signature"] as string,
      secret: env.ACTIONS_SECRET,
    });
  } catch (err) {
    return res.status(400).json({ error: "Invalid signature" });
  }

  let verdict: "allow" | "deny";

  if (context.user_data.email.split("@")[1] === "gmail.com") {
    verdict = "deny";
  } else {
    verdict = "allow";
  }

  // Sign the outgoing response using the actions secret 
  const response = await workos.actions.signResponse({
    type: "user_registration",
    verdict,
    ...(verdict === "deny" && {
      errorMessage: "Please use a work email address",
    }),
    secret: env.ACTIONS_SECRET,
  });

  return res.json(response);
});
```

## Documentation

```
[x] Yes
```

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -1,0 +1,70 @@
+import { SignatureProvider } from '../common/crypto';
+import { CryptoProvider } from '../common/crypto/crypto-provider';
+import { unreachable } from '../common/utils/unreachable';
+import {
+  AuthenticationActionResponseData,
+  ResponsePayload,
+  UserRegistrationActionResponseData,
+} from './interfaces/response-payload';
+
+export class Actions {
+  private signatureProvider: SignatureProvider;
+
+  constructor(cryptoProvider: CryptoProvider) {
+    this.signatureProvider = new SignatureProvider(cryptoProvider);
+  }
+
+  private get computeSignature() {
+    return this.signatureProvider.computeSignature.bind(this.signatureProvider);
+  }
+
+  get verifyHeader() {
+    return this.signatureProvider.verifyHeader.bind(this.signatureProvider);
+  }
+
+  serializeType(
+    type:
+      | AuthenticationActionResponseData['type']
+      | UserRegistrationActionResponseData['type'],
+  ) {
+    switch (type) {
+      case 'authentication':
+        return 'authentication_action_response';
+      case 'user_registration':
+        return 'user_registration_action_response';
+      default:
+        return unreachable(type);
+    }
+  }
+
+  async signResponse(
+    data: AuthenticationActionResponseData | UserRegistrationActionResponseData,
+    secret: string,
+  ) {
+    let errorMessage: string | undefined;
+    const { verdict, type } = data;
+
+    if (verdict === 'Deny' && data.errorMessage) {
+      errorMessage = data.errorMessage;
+    }
+
+    const responsePayload: ResponsePayload = {
+      timestamp: Date.now(),
+      verdict,
+      ...(verdict === 'Deny' &&
+        data.errorMessage && { error_message: errorMessage }),
+    };
+
+    const response = {
+      object: this.serializeType(type),
+      payload: responsePayload,
+      signature: await this.computeSignature(
+        responsePayload.timestamp,
+        responsePayload,
+        secret,
+      ),
+    };
+
+    return response;
+  }
+}
--- a/src/actions/interfaces/response-payload.ts
+++ b/src/actions/interfaces/response-payload.ts
@@ -1,0 +1,22 @@
+export interface ResponsePayload {
+  timestamp: number;
+  verdict?: 'Allow' | 'Deny';
+  errorMessage?: string;
+}
+
+interface AllowResponseData {
+  verdict: 'Allow';
+}
+
+interface DenyResponseData {
+  verdict: 'Deny';
+  errorMessage?: string;
+}
+
+export type AuthenticationActionResponseData =
+  | (AllowResponseData & { type: 'authentication' })
+  | (DenyResponseData & { type: 'authentication' });
+
+export type UserRegistrationActionResponseData =
+  | (AllowResponseData & { type: 'user_registration' })
+  | (DenyResponseData & { type: 'user_registration' });
--- a/src/common/crypto/SignatureProvider.ts
+++ b/src/common/crypto/SignatureProvider.ts
@@ -1,0 +1,76 @@
+import { SignatureVerificationException } from '../exceptions';
+import { CryptoProvider } from './CryptoProvider';
+
+export class SignatureProvider {
+  private cryptoProvider: CryptoProvider;
+
+  constructor(cryptoProvider: CryptoProvider) {
+    this.cryptoProvider = cryptoProvider;
+  }
+
+  async verifyHeader({
+    payload,
+    sigHeader,
+    secret,
+    tolerance = 180000,
+  }: {
+    payload: any;
+    sigHeader: string;
+    secret: string;
+    tolerance?: number;
+  }): Promise<boolean> {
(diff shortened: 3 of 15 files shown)


PaulAsjes (changes requested):

src/actions/actions.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  async signResponse({
+    type,
+    verdict,
+    errorMessage,
+    secret,
+  }: {
+    type: 'authentication' | 'user_registration';
+    verdict: 'allow' | 'deny';
```
tribble, line 26: I think our API interface says `Allow` or `Deny`, though we also accept the lowercase versions. We should try to be consistent. Do we prefer the capitalized version or the lowercase version?

src/actions/actions.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  get verifyHeader() {
+    return this.webhooks.verifyHeader.bind(this.webhooks);
+  }
+
+  async signResponse({
+    type,
+    verdict,
+    errorMessage,
```
tribble, line 22: Alternative suggestion to this interface. Rather than 4 parameters, just pass two: all the payload data and the secret. The difference is `errorMessage` is not part of the type in the case of an `Allow`. It also leaves us open for the response interfaces to diverge per-action in the future. Thoughts?

It would look something like this:

```
async signResponse({
  responseData, // feel free to pick a better name here
  secret
}: {
  responseData: AuthenticationActionResponseData | UserRegistrationResponseData;
  secret: string;
}) ...

interface AllowResponseData {
  verdict: 'Allow';
}

interface DenyResponseData {
  verdict 'Deny';
  errorMessage?: string; // I assume we still want this to be optional, even in the event of a Deny
}

type AuthenticationActionResponseData = 
  AllowResponseData && { type: 'authentication' }
  | DenyResponseData && { type: 'authentication' }

type UserRegistrationResponseData = 
  AllowResponseData && { type: 'user_registration' }
  | DenyResponseData && { type: 'user_registration' }
```

src/actions/actions.ts
```diff
@@ -0,0 +1,66 @@
...
+  }) {
+    const responsePayload: ResponsePayload = {
+      timestamp: Date.now(),
+      verdict,
+      ...(errorMessage && { error_message: errorMessage }),
+    };
+
+    if (verdict === 'allow' && errorMessage) {
+      throw new Error('Cannot have an error message with a verdict of allow');
+    }
```
tribble, line 38: If we change the typing like we do above this becomes less valuable... thought I guess it's still useful for anyone using plain old JS and type TypeScript?

src/actions/actions.ts
```diff
@@ -0,0 +1,66 @@
...
+      | 'user_registration_action_response';
+    switch (type) {
+      case 'authentication':
+        responseType = 'authentication_action_response';
+        break;
+      case 'user_registration':
+        responseType = 'user_registration_action_response';
+        break;
+      default:
+        throw new Error('Invalid type');
```
tribble, line 51: I wonder if we should copy over our little `unreachable` helper method. 
```
export const unreachable = (
  condition: never,
  message: string, // or make this optional
): never => {
  throw new TypeError(message);
};
```

src/actions/interfaces/response-payload.ts
```diff
@@ -0,0 +1,5 @@
+export interface ResponsePayload {
+  timestamp: number;
+  verdict?: 'allow' | 'deny';
```
tribble, line 3: Why is `verdict` optional?

tribble (commented):
Left some suggestions, let me know what you think

src/actions/actions.ts
faroceann: Great suggestion. Agree that gives us better flexibility here – it also matches the upstream API interfaces more closely.

src/actions/actions.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export class Actions {
+  private webhooks: Webhooks;
+
+  constructor(webhooks: Webhooks) {
+    this.webhooks = webhooks;
+  }
+
+  private get computeSignature() {
+    return this.webhooks.computeSignature.bind(this.webhooks);
```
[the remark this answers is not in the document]
faroceann: Went ahead with this

src/actions/actions.ts
faroceann: Let's go with capitalized here 👍

src/common/crypto/SignatureProvider.ts
```diff
@@ -0,0 +1,82 @@
+import { SignatureVerificationException } from '../exceptions';
+import { CryptoProvider } from './CryptoProvider';
+import { NodeCryptoProvider } from './NodeCryptoProvider';
+import { SubtleCryptoProvider } from './SubtleCryptoProvider';
+
+export class SignatureProvider {
```
tribble, line 6: 👍

tribble (approved):
Excited

src/actions/interfaces/response-payload.ts
```diff
@@ -0,0 +1,22 @@
...
+  errorMessage?: string;
+}
+
+interface AllowResponseData {
+  verdict: 'Allow';
+}
+
+interface DenyResponseData {
+  verdict: 'Deny';
+  errorMessage?: string; // I assume we still want this to be optional, even in the event of a Deny
```
PaulAsjes, line 13:
```suggestion
  errorMessage?: string; 
```

src/common/crypto/SignatureProvider.ts
```diff
@@ -0,0 +1,82 @@
...
+  private cryptoProvider: CryptoProvider;
+
+  constructor(subtleCrypto?: typeof crypto.subtle) {
+    if (typeof crypto !== 'undefined' && typeof crypto.subtle !== 'undefined') {
```
PaulAsjes, line 10: This unfortunately will cause problems in certain JS runtimes. Even though you're doing the check it'll still throw an error at runtime because `crypto` doesn't exist. Instead we need to use the [Node](https://github.com/workos/workos-node/blob/main/src/index.ts) and [Worker](https://github.com/workos/workos-node/blob/main/src/index.worker.ts) entry points to specify which version to use.

PaulAsjes (changes requested):

src/workos.ts
```diff
@@ -115,6 +118,10 @@ export class WorkOS {
     return new Webhooks(new SubtleCryptoProvider());
   }
 
+  createActionsClient() {
+    return new Actions(new SubtleCryptoProvider());
```
PaulAsjes, line 122: Last piece of the puzzle (and I know we're inconsistent here), you should [return an error here](https://github.com/workos/workos-node/blob/main/src/workos.ts#L130-L132) and implement `createActionsClient` in `index.ts` as well as `index.worker.ts`.

PaulAsjes (changes requested):

PaulAsjes (approved):

src/workos.ts
tribble: Sounds like there was an async convo, and we're not going to make this change.

tribble (dismissed):
One more cookie monster for you
Excited

PR #1163, opened 2024-11-14
Add access token to authentication response

PaulAsjes:
## Description
We return the access token after authenticating in the [next.js library](https://github.com/workos/authkit-nextjs?tab=readme-ov-file#get-the-access-token), this matches that behaviour for the pure Node implementation.

## Documentation

```
[ ] Yes
```


src/user-management/interfaces/authenticate-with-session-cookie.interface.ts
```diff
@@ -40,4 +40,5 @@ export type AuthenticateWithSessionCookieSuccessResponse = {
   entitlements?: string[];
   user: User;
   impersonator?: Impersonator;
+  accessToken: string;
```
mattgd, line 43: Should we add this to the docs as well?

mattgd (approved):
Looks good. Left one question inline.

PR #1176, opened 2024-11-22
fix: Update session.refresh to correctly maintain previous orgId

cmanou:
## Description

Fixes issue where:
- If you create a new session without an org
- Then refresh with an orgId 
- Then on the next refresh it no longer maintains the orgId

This was an issue in the old refreshAndSealSessionData and was fixed https://github.com/workos/workos-node/blob/08ab75f210b11371fa4de78eb97f35d2deb90a5c/src/user-management/user-management.ts#L504

Note: Having a deeper think about this, its weird that this is not required when a session is initiated with an orgId from the start which makes me think potentially there is an issue that when an refresh with orgId is sent to the workOs backend the orgId isn't actually persisted which breaks the normal refresh flow. This theory potentially also explains why in the scenario above after the switch org happens the workos portal still doesn't show the organization id on the session.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/user-management/session.ts
+++ b/src/user-management/session.ts
@@ -141,6 +141,10 @@ export class Session {
       };
     }
 
+    const { org_id: organizationIdFromAccessToken } = decodeJwt<AccessToken>(
+      session.accessToken,
+    );
+
     try {
       const cookiePassword = options.cookiePassword ?? this.cookiePassword;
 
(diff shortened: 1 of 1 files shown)


PaulAsjes (changes requested):
Thanks for this! Code looks good, but would you please rebase against main and run `npm run format` to lint the code?

PaulAsjes:
Thanks! I'll merge this now and it'll go out in the next release, which will likely be this week.

PR #1178, opened 2024-11-25
feat: introduce typed actions.deserialize

faroceann:
## Description
- Introduces typed `actions.deserialize` method 
- Export actions payload types

Usage:
```javascript

app.post("/registration-action", async (req, res) => {
  const payload = req.body;
  const sigHeader = req.headers["workos-signature"] as string;

  try {
    // context is now typed 
    const context = await workos.actions.deserialize({
      payload,
      secret: env.ACTIONS_SECRET,
      sigHeader,
      type: "user_registration",
    });

    let verdict: "Allow" | "Deny";

    if (context.user_data.email.split("@")[1] === "gmail.com") {
      verdict = "Deny";
    } else {
      verdict = "Allow";
    }

    const response = await workos.actions.signResponse(
      {
        type: "user_registration",
        verdict,
        ...(verdict === "Deny" && {
          errorMessage: "Please use a work email address",
        }),
      },
      env.ACTIONS_SECRET
    );

    return res.json(response);
  } catch (err) {
    if (err instanceof SignatureVerificationException) {
      const response = await workos.actions.signResponse(
        {
          type: "user_registration",
          verdict: "Deny",
          errorMessage: "Invalid signature",
        },
        env.ACTIONS_SECRET
      );

      return res.json(response);
    }

    const response = await workos.actions.signResponse(
      {
        type: "user_registration",
        verdict: "Deny",
        errorMessage: "An error occurred",
      },
      env.ACTIONS_SECRET
    );

    return res.json(response);
  }
});
```

## Documentation

```
[x] Yes
```

### DIFF
--- a/src/actions/actions.ts
+++ b/src/actions/actions.ts
@@ -2,6 +2,11 @@ import { SignatureProvider } from '../common/crypto';
 import { CryptoProvider } from '../common/crypto/crypto-provider';
 import { unreachable } from '../common/utils/unreachable';
 import {
+  ActionPayload,
+  AuthenticationActionPayload,
+  UserRegistrationActionPayload,
+} from './interfaces/action-payload';
+import {
   AuthenticationActionResponseData,
   ResponsePayload,
   UserRegistrationActionResponseData,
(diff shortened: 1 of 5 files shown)


src/actions/actions.ts
```diff
@@ -67,4 +72,51 @@ export class Actions {
...
+
+    switch (type) {
+      case 'authentication':
+        return payload as AuthenticationActionPayload;
+      case 'user_registration':
+        return payload as UserRegistrationActionPayload;
+      default:
+        return unreachable(type);
+    }
+  }
```
tribble, line 121: @faroceann just wondering why you took this approach with multiple method definitions?

faroceann: These overloads ensure the SDK consumer receives appropriate types based on the action type they use

tribble: @faroceann Is there something here you think I should apply to #1179?

PR #1179, opened 2024-11-25
Return action type

tribble:
## Documentation

```
[ ] Yes
```


src/actions/interfaces/action.interface.ts
```diff
@@ -0,0 +1,72 @@
...
+interface UserRegistrationAction {
+  id: string;
+  object: 'user_registration_action_context';
+  userData: UserData;
+  invitation?: Invitation;
+  ipAddress?: string;
+  userAgent?: string;
+}
+
+export type Action = AuthenticationAction | UserRegistrationAction;
```
faroceann, line 41: I prefer calling this an `ActionPayload`

src/actions/interfaces/action.interface.ts
```diff
@@ -0,0 +1,72 @@
...
+
+interface AuthenticationActionPayload {
```
faroceann, line 43: Looks like a duplicate of `AuthenticationAction`

tribble: This is the type that comes from the API response and it's snake cased. `AuthenticationAction` is what's returned to the developer and is camel cased.

package.json
```diff
@@ -9,7 +9,7 @@
     "workos"
   ],
   "volta": {
-    "node": "19.9.0",
+    "node": "20.15.0",
```
PaulAsjes, line 12: Any specific reason for the bump here?

PaulAsjes (approved):
Just one tiny nit, otherwise looks good

faroceann (approved):

package.json
tribble: Unintentional, but I couldn't run tests with the current versions due to a conflict. I can undo this.

src/actions/actions.ts
```diff
@@ -67,4 +69,21 @@ export class Actions {
 
     return response;
   }
+
+  async constructAction({
+    payload,
+    sigHeader,
+    secret,
+    tolerance = 300,
```
tribble, line 77: @faroceann calling out that I reduced this because it seems like actions should be timely... can you see any reason this would be a mistake?

faroceann: Good call out, this should be safe

PR #1193, opened 2024-12-19
Upgrade to latest iron-session

jacobwgillespie:
## Description

This PR upgrades the `iron-session` dependency from v6.3.1 to v8.0.1, for a few reasons:

1. v6.3.1 was published October 2022, over two years ago
2. v6.3.1 defines a peer dependency on `next`, which pulls in Next.js as a dependency in projects using `@workos-inc/node`, which generates GitHub vulnerability alerts for Next, even for non-Next projects

The main change to upgrade from v6 to v8 is updating the import of `iron-session/edge` to just `iron-session`, as now `iron-session` [uses the same top-level entry point](https://github.com/vvo/iron-session/releases/tag/v8.0.0) for everything. Otherwise the maintainers state that the data format should remain the same, e.g. existing sealed data should be able to be unsealed by v8.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "prepublishOnly": "yarn run build"
   },
   "dependencies": {
-    "iron-session": "~6.3.1",
+    "iron-session": "~8.0.1",
     "jose": "~5.6.3",
     "pluralize": "8.0.0"
   },
--- a/src/common/iron-session/edge-iron-session-provider.ts
+++ b/src/common/iron-session/edge-iron-session-provider.ts
@@ -1,4 +1,4 @@
-import { sealData, unsealData } from 'iron-session/edge';
+import { sealData, unsealData } from 'iron-session';
 import {
   IronSessionProvider,
   SealDataOptions,


PaulAsjes:
Hi there, thanks for this but unfortunately we can't merge this. This library needs to support Node 16, which iron-session v8.0.1 unfortunately doesn't. It's why we specify different versions depending on whether the library is [run as a worker](https://github.com/workos/workos-node/blob/main/src/index.worker.ts#L52-L54) or in a Node runtime.

jacobwgillespie:
@PaulAsjes would you ever consider inlining or replacing `iron-session`? The v6.3.1 pulls in `next`, `express`, and `koa` along with it (https://github.com/vvo/iron-session/blob/v6.3.1/package.json#L168-L172), which adds quite a bit of unnecessary dependency surface, triggering vulnerability alerts, pulling down the resolved versions of other dependencies to whatever those three agree upon, etc.

PaulAsjes:
I'd be open to inlining in this case, as long as the core functionality and API surface remained the same (plus Node 16 compatibility isn't affected).

jacobwgillespie:
Cool, I may open a PR for that - before I get too far though, how long do you expect to maintain Node 16 compatibility? As of October of last year, v16 is no longer maintained (it has exited the LTS maintenance phase). 

If you have some other reason to keep supporting v16 that's fine, but in case it's just to support Node LTS, it may be worth dropping now.

PR #1210, opened 2025-01-21
Fix library compatibility with workers environment

nicknisi:
## Description

Remove a barrel file that was importing Node-specific libraries, unnecessarily causing compatibility issues with this library when running from Cloudflare workers.

This issue first appeared in workos/authkit-nextjs#187.

## Documentation

```
[ ] Yes
```

## Testing

This wasn't caught by the `npm run test:worker` test. Manually testing with wrangler, I can see that the code fails when the `NodeCryptoProvider` is imported (via the barrel file):

capture_20250121_140633

With this change, there are no errors when running `wrangler dev`.

capture_20250121_142155


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "7.37.0",
+  "version": "7.37.1",
```
mthadley, line 2: We've generally tried to separate version bumps into their own pull-requests, instead of bundling them with other changes. For example:

https://github.com/workos/workos-node/commit/207686f530c03251d79729f41ade53f6725c3dd7

mthadley (approved):

PR #1216, opened 2025-02-05
Add permissions to roles events and API

mattgd:
## Description
Deserialize missing permissions, created at, and updated at fields on role events and add permissions to org. roles API response.

## Documentation

```
[x] Yes (API reference)
```

### DIFF
--- a/src/organizations/organizations.spec.ts
+++ b/src/organizations/organizations.spec.ts
@@ -350,7 +350,7 @@ describe('Organizations', () => {
     });
   });
 
-  describe.only('listOrganizationRoles', () => {
+  describe('listOrganizationRoles', () => {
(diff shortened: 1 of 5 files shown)


src/organizations/organizations.spec.ts
```diff
@@ -350,7 +350,7 @@ describe('Organizations', () => {
     });
   });
 
-  describe('listOrganizationRoles', () => {
+  describe.only('listOrganizationRoles', () => {
```
gcarvelli, line 353: The `.only()` was probably from testing and should get removed

gcarvelli (approved):

dangerouslycrazy:
Is this program ready to be used

PR #1236, opened 2025-03-13
Vault api

debussyman:
## Description

Adds SDK support for WorkOS Vault. Vault simplifies encryption and key management for modern web applications. It provides a service for storing sensitive data secured with scoped encryption keys, as well as a utility to encrypt data on the client using the same keys.

## Documentation

```
[X] Yes
```

[Docs PR](https://github.com/workos/workos/pull/35787)


src/common/interfaces/workos-options.interface.ts
```diff
@@ -2,8 +2,10 @@ import { AppInfo } from './app-info.interface';
 
 export interface WorkOSOptions {
   apiHostname?: string;
+  vaultHostname?: string;
```
hexedpackets, line 5: Vault specific hostname/port aren't needed with the unified routing

src/vault/interfaces/secret.interface.ts
```diff
@@ -0,0 +1,41 @@
...
+  name: string;
+  updatedAt: Date;
+}
+
+export interface SecretMetadata {
+  context: SecretContext;
+  environmentId: string;
+  id: string;
+  keyId: string;
+  updatedAt: Date;
```
hexedpackets, line 16: This is missing the `updatedBy` object containing an id and name.

That will also need to get added to `ReadSecretMetadataResponse`

src/vault/vault.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  async localEncrypt(data: string, context: SecretContext): Promise<string> {
+    const { dataKey, encryptedKey: encryptedKeys } = await this.createDataKey({
+      context,
+    });
+    return encrypt(data, dataKey.key, encryptedKeys);
+  }
+
+  async localDecrypt(encryptedData: string): Promise<string> {
```
hexedpackets, line 114: I get the point in prefixing `local` on these functions, but I think just having it as `encrypt` and `decrypt` with explanations in the docs would be more ergonomic.

debussyman:
@hexedpackets another question I had besides what `updatedBy` is for, do I need to send the `etag` in addition to the `versionId` for version checks?

src/vault/interfaces/secret/update-secret.interface.ts
```diff
@@ -1,14 +1,14 @@
...
   key_context: SecretContext;
-  version_check: string | undefined;
+  version_check?: string;
 }
 
 export interface UpdateSecretOptions {
-  name: string;
+  id: string;
   value: string;
   context: SecretContext;
```
hexedpackets, line 12: Context shouldn't be set on updates, its immutable with the current API.

debussyman: Ah nice catch!

debussyman: But the API accepts that parameter?

hexedpackets: Yeah, unknown parameters are always ignored.

src/vault/interfaces/secret.interface.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface SecretVersion {
+  createdAt: Date;
+  currentVersion: boolean;
+  id: string;
+}
+
+export interface SecretList {
+  secrets: SecretDigest[];
+  pagination: PaginationOptions;
```
hexedpackets, line 43: I think the convention is to leave the name of this key as `listMetadata`

debussyman: 👍

src/vault/vault-live-test.spec.ts
```diff
@@ -0,0 +1,284 @@
+import { disableFetchMocks, enableFetchMocks } from 'jest-fetch-mock';
+
+import { WorkOS } from '../workos';
+import { randomUUID } from 'crypto';
+import { NotFoundException } from '../index.worker';
+import { ConflictException } from '../common/exceptions/conflict.exception';
+
+describe.skip('Vault Live Test', () => {
```
mattgd, line 8: Was this test suite meant to stick around or just for development?

src/vault/serializers/vault-secret.serializer.ts
```diff
@@ -0,0 +1,81 @@
...
+} from '../interfaces';
+
+export const deserializeSecretMetadata = (
+  metadata: ReadSecretMetadataResponse,
+): SecretMetadata => ({
+  context: metadata.context,
+  environmentId: metadata.environment_id,
+  id: metadata.id,
+  keyId: metadata.key_id,
+  updatedAt: new Date(Date.parse(metadata.updated_at)),
```
mattgd, line 25: If `metadata.updated_at` is ISO-8601, this will do the trick:
```suggestion
  updatedAt: new Date(metadata.updated_at),
```

src/vault/serializers/vault-secret.serializer.ts
```diff
@@ -0,0 +1,81 @@
...
+  value: secret.value,
+  metadata: deserializeSecretMetadata(secret.metadata),
+});
+
+const deserializeSecretDigest = (
+  digest: SecretDigestResponse,
+): SecretDigest => ({
+  id: digest.id,
+  name: digest.name,
+  updatedAt: new Date(Date.parse(digest.updated_at)),
```
mattgd, line 42: Same re: date
```suggestion
  updatedAt: new Date(digest.updated_at),
```

src/vault/serializers/vault-secret.serializer.ts
```diff
@@ -0,0 +1,81 @@
...
+});
+
+export const desrializeListSecretVersions = (
+  list: ListSecretVersionsResponse,
+): SecretVersion[] => list.data.map(deserializeSecretVersion);
+
+const deserializeSecretVersion = (
+  version: SecretVersionResponse,
+): SecretVersion => ({
+  createdAt: new Date(Date.parse(version.created_at)),
```
mattgd, line 63: One more, if the date is in ISO format:
```suggestion
  createdAt: new Date(version.created_at),
```

src/vault/vault.ts
```diff
@@ -0,0 +1,135 @@
...
+    const params: string[] = [];
+    if (options?.after) {
+      params.push(`after=${encodeURIComponent(options.after)}`);
+    }
+    if (options?.limit) {
+      params.push(`limit=${options.limit}`);
+    }
+    if (params.length > 0) {
+      url = `${url}?${params.join('&')}`;
+    }
```
mattgd, line 64: No issues with this, but if you want to avoid all the string building, could do something like this:
https://github.com/workos/workos-node/blob/30efb4a16121b311ddd1909038f37e5a17d604bb/src/user-management/user-management.ts#L1040-L1050

src/workos.ts
```diff
@@ -240,7 +243,7 @@ export class WorkOS {
     console.warn(`WorkOS: ${warning}`);
   }
 
-  private handleHttpError({ path, error }: { path: string; error: unknown }) {
+  handleHttpError({ path, error }: { path: string; error: unknown }) {
```
mattgd, line 246: Any reason to remove `private` here?

mattgd (commented):
Looks good, just a few small questions and optional suggestions.

src/vault/vault-live-test.spec.ts
debussyman: I copied this pattern from FGA. It needs a running instance of Vault API, which is why I thought it was used for dev only? 
https://github.com/workos/workos-node/blob/main/src/fga/fga-live-test.spec.ts

src/vault/serializers/vault-secret.serializer.ts
debussyman: How well does that handle varying time resolution? In some instances the Vault API return millisecond resolution, in others microsecond.

src/vault/vault-live-test.spec.ts
mattgd: Ah got it, fine with me then. 👍

src/workos.ts
debussyman: That's a miss from an old version that imported the function. Will revert.

src/vault/vault.ts
debussyman: I like

src/vault/serializers/vault-secret.serializer.ts
mattgd: Good question. Microseconds...poorly. Though as far as I'm aware it would be the same with Date.parse. Pure JS is bad with microseconds in general as far as I understand, but did you find success with `new Date(Date.parse())`?

debussyman: Well it's more about being defensive. `Date.parse` will return `NaN` for formats it can't understand, while `new Date` will return a Date a epoch for formats it can't understand. I'd prefer the value to be null (likely to raise an error) than epoch (might go unnoticed). But I'm open to other approaches

debussyman: It might only be for cases of `null` - 
```
const funny = null
new Date(Date.parse(funny)) // returns null
new Date(funny) // returns epoch
```

src/workos.ts
debussyman: Done

src/vault/vault.ts
debussyman: Done

mattgd (approved):

PR #1239, opened 2025-03-18
minor version update bump

debussyman:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "7.41.0",
+  "version": "7.41.1",
   "name": "@workos-inc/node",
   "author": "WorkOS",
   "description": "A Node wrapper for the WorkOS API",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "7.41.0",
+  "version": "7.41.1",
```
amygdalama, line 2: is this for a patch (fixing a bug) or a minor version (adding a feature)? if it's a minor version it should be 7.42.0

amygdalama (approved):

package.json
nicknisi: Since this looks like it'll include #1236 it should be a minor bump (`v7.42.0`)

nicknisi (changes requested):
setting as request changes to be sure we bump the right number. Thanks!

package.json
debussyman: oh its supposed to be a minor version 🙈

PR #1247, opened 2025-03-26
Rename Vault secrets to objects

hexedpackets:
## Description

Renames `secrets` in Vault to `object`. The methods and interfaces for `secrets` are aliased to the new `object` ones to help with the transition. Eventually those will be removed completely.

## Documentation

```
[x] Yes
```

[Docs PR](https://github.com/workos/workos/pull/37306)


debussyman (commented):
Did you leave `vault.spec.ts` unchanged to test the aliasing?

hexedpackets:
Yeah exactly.

debussyman (approved):

src/vault/interfaces/secret.interface.ts
```diff
@@ -1,37 +1,16 @@
...
-  currentVersion: boolean;
-  id: string;
-}
+// tslint:disable:no-empty-interface
+export interface SecretContext extends KeyContext {}
+export interface SecretDigest extends ObjectDigest {}
+export interface SecretUpdateBy extends ObjectUpdateBy {}
+export interface SecretMetadata extends ObjectMetadata {}
+export interface VaultSecret extends VaultObject {}
+export interface SecretVersion extends ObjectVersion {}
```
mattgd, line 16: Should we mark all these as `@deprecated`?

src/vault/vault.ts
```diff
@@ -128,4 +128,13 @@ export class Vault {
...
   }
+
+  // Deprecated methods
+  createSecret = this.createObject;
+  listSecrets = this.listObjects;
+  listSecretVersions = this.listObjectVersions;
+  readSecret = this.readObject;
+  describeSecret = this.describeObject;
+  updateSecret = this.updateObject;
+  deleteSecret = this.deleteObject;
```
mattgd, line 139: Should we mark all these as `@deprecated` so it's surfaced in editors?

mattgd (approved):
Left a couple of suggestions, but looks good to me. In addition, Nick and I are passively preparing major version 8. If you'd like to put up a PR against the `v8` branch to remove the deprecated code we can make sure it gets removed in the next major version.

src/vault/vault.ts
nicknisi, line 139: suggestion: Since these are deprecated, it would be beneficial to mark them as such with a `@deprecated` tag.

```ts
/**
* @deprecated Use <OTHER_METHOD> instead
*/
```

capture_20250327_083550

nicknisi (approved):
Things look good to me. Just added one suggestion.

hexedpackets:
> Left a couple of suggestions, but looks good to me. In addition, Nick and I are passively preparing major version 8. If you'd like to put up a PR against the `v8` branch to remove the deprecated code we can make sure it gets removed in the next major version.

Awesome, I'll do that.

PR #1261, opened 2025-04-16
Add customAttributes field to SSO Profile

mattgd:
## Description
Add customAttributes field to SSO Profile.

## Documentation

```
[ ] Yes
```


src/common/interfaces/unknown-record.interface.ts
```diff
@@ -0,0 +1 @@
+export type UnknownRecord = Record<string, unknown>;
```
mattgd, line 1: Small helper type, because I was using it in a lot of places.

kendallstrautman (approved):
LGTM 👍🏻

PR #1264, opened 2025-05-01
Support optional authenticated data in local encrypt/decrypt

hexedpackets:
## Description

Add an optional parameter to the Vault local encrypt/decrypt functions for additionally authenticated data.

## Documentation

```
[x] Yes
```

https://github.com/workos/workos/pull/38690

### DIFF
--- a/src/vault/vault.ts
+++ b/src/vault/vault.ts
@@ -116,17 +116,24 @@ export class Vault {
     return deserializeDecryptDataKeyResponse(data);
   }
 
-  async encrypt(data: string, context: KeyContext): Promise<string> {
+  async encrypt(
+    data: string,
+    context: KeyContext,
+    associatedData?: string,
+  ): Promise<string> {
     const { dataKey, encryptedKeys } = await this.createDataKey({
       context,
     });
-    return encrypt(data, dataKey.key, encryptedKeys);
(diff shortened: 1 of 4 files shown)


src/vault/vault.ts
```diff
@@ -116,17 +116,24 @@ export class Vault {
...
+  async encrypt(
+    data: string,
+    context: KeyContext,
+    associatedData?: string,
+  ): Promise<string> {
     const { dataKey, encryptedKeys } = await this.createDataKey({
       context,
     });
-    return encrypt(data, dataKey.key, encryptedKeys);
+    return encrypt(data, dataKey.key, encryptedKeys, associatedData || '');
```
debussyman, line 127: Why not use the key context as the default if `aad` is not provided?

hexedpackets: 1) It would break existing ciphertext - the user would have to explicitly pass an empty string to decrypt those.
2) Serialization is doable but tricky, we would have to sort the keys or something before mapping it to a string. And make sure we keep that ordering consistent with other languages.
3) I don't think it adds anything to the encryption scheme. Since the context always maps to a single KEK, there's no additional context binding that would happen. The AAD really only helps authenticate against some non-key related data.

debussyman (approved):
You'll update the docs once all SDKs are updated?

nicknisi (approved):

PR #1270, opened 2025-05-09
[v8] Remove NodeCryptoProvider

nicknisi:
## Description

Now that we'll be supporting v18+ of Node, we no longer need the `NodeCryptoProvider`.

## Documentation

```
[ ] Yes
```


src/common/crypto/crypto-provider.spec.ts
```diff
@@ -24,7 +23,7 @@ describe('CryptoProvider', () => {
 
   describe('when computing HMAC signature', () => {
     it('returns the same for the Node crypto and Web Crypto versions', async () => {
-      const nodeCryptoProvider = new NodeCryptoProvider();
+      const nodeCryptoProvider = new SubtleCryptoProvider();
```
mattgd, line 26: Looks like we can remove any of these comparison tests now.

src/common/crypto/crypto-provider.spec.ts
```diff
@@ -46,7 +45,7 @@ describe('CryptoProvider', () => {
 
   describe('when securely comparing', () => {
     it('returns the same for the Node crypto and Web Crypto versions', async () => {
-      const nodeCryptoProvider = new NodeCryptoProvider();
+      const nodeCryptoProvider = new SubtleCryptoProvider();
```
mattgd, line 48: Looks like we can remove any of these comparison tests now.

mattgd (approved):

PR #1271, opened 2025-05-09
[v8] remove NodeHttpClient

nicknisi:
## Description

Since we're dropping support for Node 16 we can remove the Node http client, opting for the cross-runtime fetch client. This PR removes it and its usage, and drops the tests for it.

## Documentation

```
[ ] Yes
```


src/common/net/index.ts
```diff
@@ -1,19 +1,10 @@
 import { FetchHttpClient } from './fetch-client';
 import { HttpClient } from './http-client';
-import { NodeHttpClient } from './node-client';
 
 export function createHttpClient(
```
mthadley, line 4: I like leaving `createHttpClient` in place for now to make the code change easier, but maybe later we can inline `new FetchHttpClient` into the places we are calling `createHttpClient` and get rid of that layer of indirection.

mthadley: Wait, do we even call this anywhere? I see down below we had this same logic under `WorkOSNode.createHttpClient` method. 🤔

In any case, something that can be cleaned up later.

mthadley (approved):

src/common/net/index.ts
nicknisi: Yeah, you're right. This file isn't used at all and can be removed.

PR #1280, opened 2025-05-19
Add warnings to FGA query and check responses

atainter:
## Description
Adds warnings field to the FGA check and query responses. Warning types include:

BaseWarning - catch all for warnings not supported by the current SDK version
MissingContextKeysWarning - missing context keys found while evaluating policies in the check / query request

Note: 

Query response returns "warnings" as a top-level field in the Paginate response. To avoid affecting other packages, I extended the AutoPaginate class in FGA package and added a custom deserializer.


stanleyphu (approved):

src/fga/utils/fga-paginatable.ts
```diff
@@ -0,0 +1,8 @@
+import { AutoPaginatable } from '../../common/utils/pagination';
+import { Warning } from '../interfaces/warning.interface';
+
+export class FgaPaginatable<T> extends AutoPaginatable<T> {
+  get warnings(): Warning[] | undefined {
+    return (this as any).list.warnings;
```
nicknisi, line 6: **question(non-blocking):** Would it be better DX to always return an array rather than an array or `undefined`?

```diff
-  get warnings(): Warning[] | undefined {
+  get warnings(): Warning[] {
-    return (this as any).list.warnings;
+    return (this as any).list.warnings ?? [];
```

nicknisi: Also, wish we weren't casting `this` as `any` but looking at this it might be more involved than just extending the generic.

nicknisi (approved):
No blockers from me!

src/fga/utils/fga-paginatable.ts
atainter: > Would it be better DX to always return an array rather than an array or undefined
Possibly, but it's consistent with the API / other SDKs where the server response omits the key if empty. What do you think?

nicknisi: Yep, not blocking on my part and if it's more consistent I get it.

PR #1307, opened 2025-07-11
Type constructEvent payload parameter as Record<string, unknown>

mattgd:
## Description
Type constructEvent payload parameter as Record<string, unknown>.

Example usage:
```
app.post("/webhooks", async (request, response) => {
  const payload = request.body;
  const sigHeader = request.headers["workos-signature"] as string;

  const webhook = await workosClient.webhooks.constructEvent({
    payload,
    sigHeader,
    secret: WEBHOOK_SECRET,
  });

  const data = webhook.data;

  // Verify the signature and process the event
  response.status(200).end();
});
```

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/webhooks/webhooks.ts
+++ b/src/webhooks/webhooks.ts
@@ -30,7 +30,7 @@ export class Webhooks {
     secret,
     tolerance = 180000,
   }: {
-    payload: unknown;
+    payload: Record<string, unknown>;
     sigHeader: string;
     secret: string;
     tolerance?: number;
@@ -38,7 +38,7 @@ export class Webhooks {
     const options = { payload, sigHeader, secret, tolerance };
     await this.verifyHeader(options);
 
-    const webhookPayload = payload as EventResponse;
+    const webhookPayload = payload as unknown as EventResponse;
 
     return deserializeEvent(webhookPayload);
   }


src/workos.ts
```diff
@@ -66,7 +66,10 @@ export class WorkOS {
   readonly widgets = new Widgets(this);
   readonly vault = new Vault(this);
 
-  constructor(readonly key?: string, readonly options: WorkOSOptions = {}) {
+  constructor(
+    readonly key?: string,
+    readonly options: WorkOSOptions = {},
+  ) {
```
mattgd: Not sure why CI was giving me a formatter error for this.

src/sso/interfaces/authorization-url-options.interface.ts
```diff
@@ -8,9 +8,3 @@ export interface SSOAuthorizationURLOptions {
   redirectUri: string;
   state?: string;
 }
-
-/**
- * @deprecated Use SSOAuthorizationURLOptions instead
- */
-// tslint:disable-next-line:no-empty-interface
-export interface AuthorizationURLOptions extends SSOAuthorizationURLOptions {}
```
mattgd: Was getting a linter error in CI for this, but I assume it should just be removed.

PR #1308, opened 2025-07-16
add support for `provider_query_params` option on getAuthorizationUrl

stacurry:
## Description
- moves to using `qs` to create query strings for `getAuthorizationUrl` to more easily support a new object type parameter and stringify it into the format our API expects
  - we have lots of tests for `getAuthorizationUrl` (sso and user_management) that compare full url strings to snapshots, so the fact that all of these tests still pass (with `qs` options set to maintain backwards compatibility with our previous implementation) makes this feel pretty safe to me, but if anyone disagrees, let me know
- adds support for the new `provider_query_params` query parameter / option added in https://github.com/workos/workos/pull/41556

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/user-management/__snapshots__/user-management.spec.ts.snap
+++ b/src/user-management/__snapshots__/user-management.spec.ts.snap
@@ -12,6 +12,8 @@ exports[`UserManagement getAuthorizationUrl with a provider generates an authori
 
 exports[`UserManagement getAuthorizationUrl with a provider with providerScopes generates an authorize url that includes the specified scopes 1`] = `"https://api.workos.com/user_management/authorize?client_id=proj_123&provider=GoogleOAuth&provider_scopes=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcalendar&provider_scopes=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fadmin.directory.group&redirect_uri=example.com%2Fauth%2Fworkos%2Fcallback&response_type=code"`;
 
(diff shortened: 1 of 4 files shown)


nholden (approved):

package.json
```diff
@@ -41,7 +41,8 @@
     "iron-session": "~6.3.1",
     "jose": "~5.6.3",
     "leb": "^1.0.0",
-    "pluralize": "8.0.0"
+    "pluralize": "8.0.0",
+    "qs": "6.14.0"
```
jonatascastro12, line 45: I'd avoid adding the new package `qs` here if it's possible to use the native `URLSearchParams` (to avoid extra packages installed in the SDK).

stacurry: the native `URLSearchParams` doesn't have good support for array or object parameters, like `provider_scopes` and now `provider_query_params`. writing our own logic for array params was ok, but I was worried about missing edge cases if we tried to write our own logic to handle objects so I thought this was worth the added dependency.

MTG2000:
Hey @stacurry 

Not sure if this is a bug or not, but I noticed that after updating to this version, the authorization urls generated with a query parameter with `undefined` value are now resulting in broken urls.

So generating a url previously like this:
```
workos.userManagement.getAuthorizationUrl({
      provider: 'authkit',
      clientId: this.clientId,
      redirectUri: callbackUrl,
      state: options.state,
      screenHint: undefined,
    })
 ```
 Would result in a url like this: (where the `screenHint` param has been removed)
 ```
https://api.workos.com/user_management/authorize?client_id=client_0000&provider=authkit&redirect_uri=https%3A%2F%2Fapp.something.com%2F_auth%2Fcallback&response_type=code
```

Whereas now, the same call would return this URL: (note the empty `screen_hint` at the end)
```
https://api.workos.com/user_management/authorize?client_id=client_000&provider=authkit&redirect_uri=https%3A%2F%2Fapp.something.com%2F_auth%2Fcallback&response_type=code&screen_hint=
```
Which results in an error page.

I fixed it temporarily by conditionally passing the `screenHint` value,
```
workos.userManagement.getAuthorizationUrl({
      provider: 'authkit',
      clientId: this.clientId,
      redirectUri: callbackUrl,
      ...(options.state && {
        state: options.state,
      }),
      ...(options.signInType && {
        screenHint: options.signInType,
      }),
    })
 ```
 
But if this new behaviour is intended, it should have been marked as a breaking change I think.

PR #1316, opened 2025-07-22
Add `userManagement.listSessions`

cmatheson:
## Description

This endpoint will list the active sessions for a given user.

## Documentation

```
[X] Yes
```


src/common/interfaces/event.interface.ts
```diff
@@ -22,8 +22,8 @@ import {
   MagicAuthEventResponse,
   PasswordResetEvent,
   PasswordResetEventResponse,
-  Session,
```
cmatheson: i hate this rename, but we've got a conflict with the `Session` class (that one probably should have been called CookieSession or something?).

cmatheson: this is possibly a breaking change?

nicknisi: Yeah, unfortunately I think this is a breaking change.

Could we do something like this for now and then clean it up in the `version-8` branch?

```ts
  /**
   * @deprecated Use UserSession instead. Will be removed in v8.0.0
   */
  export type Session = UserSession;

  /**
   * @deprecated Use UserSessionResponse instead. Will be removed in v8.0.0
   */
  export type SessionResponse = UserSessionResponse;
```

nicknisi (changes requested):
LGTM other than the breaking change

src/common/interfaces/event.interface.ts
cmatheson: decided to rename the (class) Session -> CookieSession. i believe this is a non-breaking change (we don't export the Session class directly).

nicknisi (approved):

PR #1317, opened 2025-07-23
Allow passing a known key to Vault encrypt/decrypt functions

hexedpackets:
## Description

This lets an application cache a data key and use it for multiple operations without having to call the Vault API each time.

## Documentation

```
[ ] Yes
```

### DIFF
--- a/src/vault/vault.ts
+++ b/src/vault/vault.ts
@@ -52,7 +52,7 @@ export class Vault {
     this.cryptoProvider = workos.getCryptoProvider();
   }
 
-  private decode(payload: string): Decoded {
+  async decode(payload: string): Promise<Decoded> {
     const inputData = base64ToUint8Array(payload);
     // Use 12 bytes for IV (standard for AES-GCM)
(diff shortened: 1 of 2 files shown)


src/vault/vault.ts
```diff
@@ -213,11 +216,18 @@ export class Vault {
...
   async decrypt(
-    encryptedData: string,
+    encryptedData: string | Decoded,
     associatedData?: string,
+    dataKey?: DataKey,
   ): Promise<string> {
-    const decoded = this.decode(encryptedData);
-    const dataKey = await this.decryptDataKey({ keys: decoded.keys });
+    const decoded =
+      typeof encryptedData === 'string'
```
debussyman, line 224: This feels strange to me, like two methods are being mixed here. Why not introduce localencrypt/localdecrypt methods that have distinct signatures from encrypt/decrypt?

debussyman:
@gjtorikian We should close out this PR and revisit the implementation

PR #1337, opened 2025-08-11
add `widgets:sso:manage` scope

cmatheson:
## Description

Adds the `widgets:sso:manage` scope. Also makes `userId` optional (We anticipate customer's using this widget without a corresponding AuthKit user).

## Documentation

```
[X] Yes
```


src/widgets/interfaces/get-token.ts
```diff
@@ -1,14 +1,14 @@
-export type WidgetScope = 'widgets:users-table:manage';
+export type WidgetScope = 'widgets:users-table:manage' | 'widgets:sso:manage';
 
 export interface GetTokenOptions {
   organizationId: string;
-  userId: string;
+  userId?: string;
```
cmatheson, line 5: considered doing something fancy here where `userId` is conditionally required depending on what `WidgetScope`s are requested, but it seemed overly complicated.

mthadley (approved):

PR #1345, opened 2025-08-22
Add `providerScopes` and `oauthTokens` support to SSO API

jonatascastro12:
## Summary
- Add `providerScopes` parameter to SSO authorization URL options
- Add `oauthTokens` field to SSO profile and token responses
- Comprehensive test coverage for new functionality

## Changes Made
- Updated `SSOAuthorizationURLOptions` interface to include optional `providerScopes` parameter
- Modified `getAuthorizationUrl` method to serialize provider scopes as space-separated string
- Added `oauthTokens` field to `ProfileAndToken` and `ProfileAndTokenResponse` interfaces  
- Updated profile and token serializer to handle oauth tokens deserialization
- Added tests for provider scopes in authorization URLs
- Added tests for oauth tokens in profile responses

## Test Plan
- [x] All existing tests pass
- [x] New tests added for provider scopes functionality
- [x] New tests added for oauth tokens functionality
- [x] Linter passes
- [x] TypeScript compilation successful

Related to ENT-3768

### DIFF
--- a/src/sso/interfaces/profile-and-token.interface.ts
+++ b/src/sso/interfaces/profile-and-token.interface.ts
@@ -1,5 +1,8 @@
 import { UnknownRecord } from '../../common/interfaces/unknown-record.interface';
-import { OauthTokens, OauthTokensResponse } from '../../user-management/interfaces/oauth-tokens.interface';
+import {
+  OauthTokens,
+  OauthTokensResponse,
+} from '../../user-management/interfaces/oauth-tokens.interface';
 import { Profile, ProfileResponse } from './profile.interface';
 
 export interface ProfileAndToken<CustomAttributesType extends UnknownRecord> {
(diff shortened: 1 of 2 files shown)


nicknisi (approved):

src/sso/sso.ts
```diff
@@ -93,6 +94,7 @@ export class SSO {
       domain_hint: domainHint,
       login_hint: loginHint,
       provider,
+      provider_scopes: providerScopes?.join(' '),
```
stacurry, line 97: I would just pass in `providerScopes` here like we do for user management and then `toQueryString` will take care of serializing. The API is expecting a comma-separated list or a repeated param here so I don't think a space separated list will work as expected.

jonatascastro12: Will patch a fix

PR #1351, opened 2025-08-27
Add timeout to http client requests

nave91:
## Description
`fetch` does not have a default timeout, so there is a non-zero chance of our HTTP requests to WorkOS servers hanging indefinitely. This PR introduces `timeout` as an optional parameter while initializing WorkOS client to terminate the request if its taking longer than `timeout` milliseconds. 

## Documentation

```
[ ] Yes
```


nave91:
> Should we have a default _conservative_ timeout? Maybe `30s`?

Added a default of 60s to be very conservative

PR #1355, opened 2025-09-08
Export session error enums for type-safe error handling

### LINKED ISSUE #959: Types for errors
It seems like when the internal works api (api.workos.com) throws an error you just return it.
For example the api `authenticateWithCode` can throw an error for email verification, but the types for the errors are not in this library.

Can you guys please add a custom error type and interfaces for your errors?

Here is how I need to handle it now:
```ts
		return workos.userManagement
			.authenticateWithCode({clientId, code,})
			.catch((error) => {
				// The error code should be as described here:
				// https://workos.com/docs/reference/user-management/authentication-errors
				if (error.code === "email_verification_required") {
					context.set.status = 401; // unauthorized
					return error;
				}

                                // Additional if clauses for other errors.

				// Else let the error bubble up
				throw error;
			});
```
Now I have two problems, a) my API is not typed unless I create types for all the errors manually, and b) I need to write code to handle all the workOS errors to separate them from server errors, as can be seen I the `if` clause.

If you were to use custom error classes then problem "a" would be solved, and if you add types then problem "b" would be solved. Or better yet you could create both and type the classes so that we get the best of bth worlds.

My code would then look like this:

```ts
		return workos.userManagement
			.authenticateWithCode({clientId, code,})
			.catch((error) => {
			        if (error instanceof AuthenticationError) {
			          // handle works errors
			        }
			        
				// Else let the error bubble up
				throw error;
			});
```

And the error will be typed!

Mearman:
## Summary
Exports `RefreshAndSealSessionDataFailureReason` and `AuthenticateWithSessionCookieFailureReason` enums along with related session types to enable type-safe error handling.

## Changes
- Export session error enums from user-management interfaces
- Add comprehensive tests for exported enums

## Addresses
Resolves #959 - Provides access to error codes like `MFA_ENROLLMENT`, `INVALID_GRANT`, and `INVALID_SESSION_COOKIE` that developers requested for proper error classification.

### DIFF
--- a/src/user-management/error-enums.spec.ts
+++ b/src/user-management/error-enums.spec.ts
@@ -1,0 +1,53 @@
+import { 
+  RefreshAndSealSessionDataFailureReason, 
+  AuthenticateWithSessionCookieFailureReason 
+} from './interfaces';
+
+describe('WorkOS Error Enums - Minimal Export Fix', () => {
+  it('should export RefreshAndSealSessionDataFailureReason enum', () => {
+    // These are the error codes WorkOS already uses internally
+    expect(RefreshAndSealSessionDataFailureReason.INVALID_SESSION_COOKIE).toBe('invalid_session_cookie');
+    expect(RefreshAndSealSessionDataFailureReason.NO_SESSION_COOKIE_PROVIDED).toBe('no_session_cookie_provided');
+    expect(RefreshAndSealSessionDataFailureReason.INVALID_GRANT).toBe('invalid_grant');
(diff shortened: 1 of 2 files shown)


Mearman:
Closing this PR as the error enums are already exported in the current codebase. The enums RefreshAndSealSessionDataFailureReason and AuthenticateWithSessionCookieFailureReason have been available since they were added to the interfaces. The issue appears to be one of discoverability rather than missing exports. Developers can already import these enums using:

```typescript
import { RefreshAndSealSessionDataFailureReason, AuthenticateWithSessionCookieFailureReason } from '@workos-inc/node';
```

A documentation update would be more appropriate to help developers discover and use these existing error handling capabilities.

PR #1376, opened 2025-10-29
Bump to v7.72.1

hexedpackets:
## Documentation

```
[ ] Yes
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "7.72.0",
+  "version": "7.72.1",
   "name": "@workos-inc/node",
   "author": "WorkOS",
   "description": "A Node wrapper for the WorkOS API",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "7.72.0",
+  "version": "7.72.1",
```
nicknisi, line 2: I ran locally and it made changes to `package-lock.json`. Could you check in those changes, too?

nicknisi (changes requested):

package.json
hexedpackets: Done. FYI I typically use `yarn` which doesn't modify `package-lock.json`, and `npm run build` also doesn't change it. I had to call `npm install` to bump the lock file.

PR #1416, opened 2025-12-12
[v8] Replace iron-session with iron-webcrypto v2

nicknisi:
## Summary
- Replaces `iron-session` dependency with direct `iron-webcrypto` v2.0.0
- Creates lightweight `seal.ts` wrapper providing iron-session compatible API
- Reduces dependency footprint while maintaining backwards compatibility

## Changes
- Add `src/common/crypto/seal.ts` with `sealData` and `unsealData` functions
- Update `package.json` to use `iron-webcrypto` ^2.0.0
- Update `jest.config.cjs` to transform ESM-only `uint8array-extras` dependency
- Update imports in session and user-management modules

### DIFF
--- a/src/common/crypto/seal.ts
+++ b/src/common/crypto/seal.ts
@@ -9,12 +9,10 @@ const CURRENT_MAJOR_VERSION = 2;
 
 interface SealOptions {
   password: string;
-  ttl?: number;
 }
 
 interface UnsealOptions {
   password: string;
-  ttl?: number;
 }
 
 function parseSeal(seal: string): {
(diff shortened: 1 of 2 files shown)


src/common/crypto/seal.ts
```diff
@@ -0,0 +1,84 @@
...
+  const { sealWithoutVersion, tokenVersion } = parseSeal(encryptedData);
+
+  const passwordMap = { 1: password };
+
+  let data: unknown;
+  try {
+    data =
+      (await ironUnseal(sealWithoutVersion, passwordMap, {
+        ...defaults,
+        ttl: ttl * 1000,
```
cmatheson: we don't really use `ttl` do we? i'd be ok with just implementing the interface we need if you want to drop that.

nicknisi: Good catch. I'll simplify the interface to just password.

PR #1420, opened 2025-12-16
Add support for feature flags management API

stanleyphu:
## Description

- Adds the following methods to support the feature flags management API
  - `listFeatureFlags` (`GET /feature-flags`)
  - `getFeatureFlag` (`GET /feature-flags/:slug`)
  - `enableFeatureFlag` (`PUT /feature-flags/:slug/enable`)
  - `disableFeatureFlag` (`PUT /feature-flags/:slug/disable`)
  - `addFlagTarget` (`POST /feature-flags/:slug/targets/:targetId`)
  - `removeFlagTarget` (`DELETE /feature-flags/:slug/targets/:targetId`)
- Updates `FeatureFlag` interface to add `tags`, `enabled` and `defaultValue` fields
- Updates `FeatureFlag.description` to be a required field to match the API response

## Documentation

```
[x] Yes
```

### DIFF
--- a/src/feature-flags/feature-flags.ts
+++ b/src/feature-flags/feature-flags.ts
@@ -1,9 +1,11 @@
 import { AutoPaginatable } from '../common/utils/pagination';
 import { WorkOS } from '../workos';
 import {
+  AddFlagTargetOptions,
   FeatureFlag,
   FeatureFlagResponse,
   ListFeatureFlagsOptions,
+  RemoveFlagTargetOptions,
 } from './interfaces';
 import { deserializeFeatureFlag } from './serializers';
 import { fetchAndDeserialize } from '../common/utils/fetch-and-deserialize';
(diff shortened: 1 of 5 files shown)


src/feature-flags/feature-flags.ts
```diff
@@ -0,0 +1,68 @@
...
+    return deserializeFeatureFlag(data);
+  }
+
+  async addFlagTarget(slug: string, targetId: string): Promise<void> {
+    await this.workos.post(`/feature-flags/${slug}/targets/${targetId}`, {});
+  }
+
+  async removeFlagTarget(slug: string, targetId: string): Promise<void> {
+    await this.workos.delete(`/feature-flags/${slug}/targets/${targetId}`);
+  }
```
nholden: Did you consider making `slug` and `targetId` named arguments instead of positional arguments? The order of those arguments aren't particularly obvious, so I'm wondering if `addFlagTarget({ slug: 'my-feature', targetId: 'user_foobar' })` might be more foolproof.

stanleyphu: I agree - I think named arguments makes more sense here. Updated the PR to do this

PR #1435, opened 2026-01-05
[v8] feat: add PKCE support for public clients; remove /client entry point

nicknisi:
## Description

Enable PKCE authentication for both public and confidential clients.

### Changes

- **API key now optional**: Initialize with just `clientId` for PKCE mode: `new WorkOS({ clientId: 'client_...' })`
- **New helper method**: `getAuthorizationUrlWithPKCE()` - generates PKCE internally, returns `{ url, state, codeVerifier }`
- **Enhanced exchange**: `authenticateWithCode()` auto-detects client mode based on available credentials
- **Manual PKCE option**: `workos.pkce.generate()` + `getAuthorizationUrl()` for advanced use cases
- **Non-breaking**: Existing `getAuthorizationUrl()` unchanged, still returns URL string

### PKCE with Confidential Clients (OAuth 2.1 Best Practice)

Server-side apps can use PKCE alongside the client secret for defense in depth:

```ts
const workos = new WorkOS('sk_...'); // With API key

const { url, codeVerifier } = await workos.userManagement.getAuthorizationUrlWithPKCE({
  provider: 'authkit',
  redirectUri: 'https://example.com/callback',
  clientId: 'client_...',
});

// Both client_secret AND code_verifier will be sent
const { accessToken } = await workos.userManagement.authenticateWithCode({
  code: authorizationCode,
  codeVerifier,
  clientId: 'client_...',
});
```

The auto-detection logic:
| API Key | codeVerifier | Behavior |
|---------|--------------|----------|
| ✅ | ✅ | Send both `client_secret` AND `code_verifier` (confidential + PKCE) |
| ✅ | ❌ | Send `client_secret` only (traditional confidential client) |
| ❌ | ✅ | Send `code_verifier` only (public client) |
| ❌ | ❌ | Error |

### Removed: `@workos-inc/node/client` export

The separate `/client` entry point has been removed. Instead of:

```ts
// Old approach - standalone functions
import { userManagement } from '@workos-inc/node/client';
const url = userManagement.getAuthorizationUrl({ ... });
```

Use the standard SDK without an API key:

```ts
// New approach - consistent with rest of SDK
import { WorkOS } from '@workos-inc/node';
const workos = new WorkOS({ clientId: 'client_...' });
const url = workos.userManagement.getAuthorizationUrl({ ... });
```

This provides a single, consistent API surface rather than two parallel approaches.

### Public Client Usage

```ts
import { WorkOS } from '@workos-inc/node';

const workos = new WorkOS({ clientId: 'client_...' });

// Step 1: Get authorization URL with auto-generated PKCE
const { url, state, codeVerifier } = await workos.userManagement.getAuthorizationUrlWithPKCE({
  redirectUri: 'myapp://callback',
  provider: 'authkit',
});

// Store codeVerifier securely, then redirect user to url

// Step 2: Exchange code for tokens
const { accessToken, refreshToken, user } = await workos.userManagement.authenticateWithCode({
  code: authCode,
  codeVerifier,
});
```

### Methods Available Without API Key

| Method | Description |
|--------|-------------|
| `userManagement.getAuthorizationUrlWithPKCE()` | Build OAuth URL with auto-generated PKCE |
| `userManagement.getAuthorizationUrl()` | Build OAuth URL (with manual PKCE params) |
| `userManagement.authenticateWithCode()` | Exchange code + verifier for tokens |
| `userManagement.authenticateWithCodeAndVerifier()` | Exchange code + verifier for tokens (explicit) |
| `userManagement.authenticateWithRefreshToken()` | Refresh tokens |
| `userManagement.getLogoutUrl()` | Build logout redirect URL |
| `userManagement.getJwksUrl()` | Get JWKS URL for local JWT validation |
| `workos.pkce.generate()` | Generate PKCE code verifier and challenge |

### DIFF
--- a/src/user-management/interfaces/authenticate-with-options-base.interface.ts
+++ b/src/user-management/interfaces/authenticate-with-options-base.interface.ts
@@ -36,4 +36,6 @@ export interface SerializedAuthenticatePublicClientBase {
  * const serialize = (options: WithResolvedClientId<SomeOptions>) => ...
  * ```
  */
-export type WithResolvedClientId<T> = Omit<T, 'clientId'> & { clientId: string };
+export type WithResolvedClientId<T> = Omit<T, 'clientId'> & {
+  clientId: string;
+};


src/index.ts
```diff
@@ -26,18 +26,33 @@ export * from './roles/interfaces';
...
   /** @override */
   createHttpClient(options: WorkOSOptions, userAgent: string): HttpClient {
+    const headers: Record<string, string> = {
+      'User-Agent': userAgent,
+    };
+
+    const configHeaders = options.config?.headers;
+    if (
+      configHeaders &&
+      typeof configHeaders === 'object' &&
```
cmatheson: i'm a little confused about what these extra checks are for. do we need to do something else with the headers if they are an array or `instanceof Headers`?)

src/workos.ts
```diff
@@ -74,18 +80,38 @@ export class WorkOS {
...
+   * @param keyOrOptions - API key string, or options object for PKCE/public clients
+   * @param maybeOptions - Options when first argument is API key
+   *
+   * @example
+   * // Server-side with API key
+   * const workos = new WorkOS('sk_...');
+   *
+   * @example
+   * // PKCE/public client (no API key)
+   * const workos = new WorkOS({ clientId: 'client_...' });
```
cmatheson, line 99: i wonder if we could keep the API key mode around for backwards compat, but prefer/encourage the object mode going forward. i'm imagining that even confidential clients would initialize with:

```
const workos = new WorkOS({
  client_id: 'client_1234',
  client_secret: 'sssshhh'
});
```

It feels strange to me to initialize differently based on how you want to authenticate.

I would also really love it if the authkit APIs could use the clientId specified here for all authenticate calls instead of awkwardly requiring the user to specify them as arguments. (in authenticateWithPassword/getJwksUrl/etc.). (we don't necessarily need to tackle that right now though).

src/workos.ts
```diff
@@ -74,18 +80,38 @@ export class WorkOS {
...
+    } else {
+      this.key = keyOrOptions;
+      this.options = maybeOptions ?? {};
+    }
 
-      if (!this.key) {
-        throw new NoApiKeyProvidedException();
-      }
+    if (!this.key) {
+      this.key = getEnv('WORKOS_API_KEY');
```
cmatheson, line 114: i kind of think it would be lovely if you could specify your WORKOS_CLIENT_ID via env var too.

cmatheson: (i really wish client_id was required but i also see the value of having it be optional to preserve backwards compatibility).

cmatheson: also wondering if we could only attach attach the stuff that makes sense for how this was initialized (e.g. if you don't specify a client_secret, we don't attach all the CRUD objects that you currently will get a run-time error (no key) for when calling). i think this would make the typescript experience pretty awesome.

README.md
```diff
@@ -37,6 +37,54 @@ import { WorkOS } from '@workos-inc/node';
...
+Server-side apps can also use PKCE alongside the client secret for defense in depth (recommended by OAuth 2.1):
+
+```ts
+const workos = new WorkOS('sk_...'); // With API key
+
+// Use PKCE even with API key for additional security
+const { url, codeVerifier } = await workos.userManagement.getAuthorizationUrlWithPKCE({
+  provider: 'authkit',
+  redirectUri: 'https://example.com/callback',
+  clientId: 'client_...',
```
cmatheson: this is the redundant clientId stuff that i wish we could omit given we already specified this when creating the client.

README.md
```diff
@@ -37,6 +37,54 @@ import { WorkOS } from '@workos-inc/node';
...
+
+For apps that can't securely store secrets, initialize with just a client ID:
+
+```ts
+import { WorkOS } from '@workos-inc/node';
+
+const workos = new WorkOS({ clientId: 'client_...' }); // No API key needed
+
+// Generate auth URL with automatic PKCE
+const { url, codeVerifier } = await workos.userManagement.getAuthorizationUrlWithPKCE({
```
cmatheson: should we mention that the codeVerifier should be stored on-device (cookie, phone storage, etc.) for later? (it's somewhat clear since you need it for the authenticateWithCode call, but maybe not obvious to someone who has never done this before?

src/index.ts
nicknisi: Yeah, I meant to handle those other types for headers and must've gotten distracted. I fixed it.

src/workos.ts
nicknisi: > i kind of think it would be lovely if you could specify your WORKOS_CLIENT_ID via env var too.

This is already [supported](https://github.com/workos/workos-node/blob/6f168630e72ba845181f71ffd2100619c835922e/src/workos.ts#L120-L122).

nicknisi: > also wondering if we could only attach attach the stuff that makes sense for how this was initialized (e.g. if you don't specify a client_secret, we don't attach all the CRUD objects that you currently will get a run-time error (no key) for when calling). i think this would make the typescript experience pretty awesome.

I don't disagree here and shied away from it because it'd be a pretty big breaking change. Constructors can't be overloaded in TS I don't think, so the best way to apply this would probably be with a factory function. That's something we could look at adding later, but for now users will get a runtime error if using a method that requires an API Key when none is provided.

README.md
nicknisi: `clientId` is now optional in all method parameters. If not provided, it falls back to the value from the constructor (`new WorkOS('sk_...', { clientId: 'client_123' })`).

Changes:
- Added `resolveClientId()` helper that throws a clear error if `clientId` isn't available from either source
- Updated all 10 affected methods (8 authenticate methods + 2 authorization URL methods)
- Added WithResolvedClientId<T> utility type for type-safe serializers (no ! assertions)
- Added tests for fallback behavior

Usage:
```ts
const workos = new WorkOS('sk_...', { clientId: 'client_123' });

// clientId now optional - falls back to constructor value
await workos.userManagement.authenticateWithCode({ code });
```

README.md
nicknisi: Added the following to the README:

> [!IMPORTANT]
> Store `codeVerifier` securely on-device between generating the auth URL and handling the callback. For mobile apps, use platform secure storage (iOS Keychain, Android Keystore). For CLI apps, consider OS credential storage. The verifier must survive app restarts during the auth flow.

PR #1440, opened 2026-01-08
feat: add createWorkOS factory for type-safe public/confidential clients [v8]

nicknisi:
## Summary

Adds a `createWorkOS()` factory function that provides **compile-time type safety** for public vs confidential clients.

When the SDK is instantiated with only a `clientId` (no API key), most methods throw `ApiKeyRequiredException` at runtime. However, TypeScript can't warn you at compile time because the `WorkOS` class type is static—all methods appear available regardless of how it was constructed.

The factory solves this by using function overloads that return different types based on input:

- `createWorkOS({ clientId })` → returns `PublicWorkOS` (narrow type exposing only PKCE-compatible methods)
- `createWorkOS({ apiKey, ... })` → returns full `WorkOS` type

## Example

```typescript
import { createWorkOS } from '@workos-inc/node';

// Public client (Electron, mobile, CLI) - only PKCE methods available
const publicClient = createWorkOS({ clientId: 'client_123' });

// ✅ These work - available on PublicWorkOS
const { url, codeVerifier } = await publicClient.userManagement.getAuthorizationUrlWithPKCE({
  provider: 'authkit',
  redirectUri: 'myapp://callback',
});

const auth = await publicClient.userManagement.authenticateWithCodeAndVerifier({
  code: authCode,
  codeVerifier,
});

// ❌ TypeScript error - not available on PublicWorkOS
publicClient.userManagement.listUsers();
publicClient.organizations.list();

// Confidential client (server) - full access
const serverClient = createWorkOS({
  apiKey: process.env.WORKOS_API_KEY!,
  clientId: 'client_123',
});

// ✅ All methods available
await serverClient.userManagement.listUsers();
await serverClient.organizations.list();
```

## Design

- **No runtime changes** - The factory returns a standard `WorkOS` instance; type narrowing is purely compile-time
- **Uses `Pick<>` for maintainability** - Public method names are listed once as a string union; types are derived automatically
- **Ignores env vars** - Factory uses only explicit input for predictable types. Users who want env var convenience can use `new WorkOS()` or pass `process.env.WORKOS_API_KEY` explicitly
- **Backward compatible** - `new WorkOS()` continues to work unchanged

### DIFF
--- a/src/factory.spec.ts
+++ b/src/factory.spec.ts
@@ -86,22 +86,6 @@ describe('createWorkOS', () => {
     });
   });
 
-  describe('with string apiKey', () => {
-    it('returns a WorkOS instance', () => {
-      const workos = createWorkOS('sk_test_123');
-
-      expect(workos.key).toBe('sk_test_123');
-      expect(workos.baseURL).toBeDefined();
-    });
-
-    it('accepts additional options', () => {
-      const workos = createWorkOS('sk_test_123', { clientId: 'client_123' });
-
-      expect(workos.key).toBe('sk_test_123');
-      expect(workos.clientId).toBe('client_123');
-    });
-  });
-
   /**
    * Type-level tests - these verify compile-time behavior.
    * If these compile, the types are working correctly.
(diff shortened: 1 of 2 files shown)


src/factory.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+
+describe('createWorkOS', () => {
+  const OLD_ENV = process.env;
+
+  beforeEach(() => {
+    jest.resetModules();
+    process.env = { ...OLD_ENV };
+    delete process.env.WORKOS_API_KEY;
+    delete process.env.WORKOS_CLIENT_ID;
+  });
```
imkesin, line 15: If we want to be sure that `createWorkOS` doesn't rely on these, there could be some approach with a `Proxy` to throw an error when we try to get certain variables.

(Obviously would be more work + gross)

nicknisi: Yeah IMO it's not worth it. This give you the compile-time safety but if you have the env var set it'll still work but at that point you're ignoring TS errors or using JS and not worried about the factory function's return type, anyway.

PR #1454, opened 2026-01-29
feat: Add environment role support

csrbarber:
## Description

- Adds PATCH request support
- Adds Environment Role support

## Documentation

```
[X] Yes
```

### DIFF
--- a/src/authorization/authorization.ts
+++ b/src/authorization/authorization.ts
@@ -7,6 +7,8 @@ import {
   CreateEnvironmentRoleOptions,
   UpdateEnvironmentRoleOptions,
   ListEnvironmentRolesOptions,
+  SetEnvironmentRolePermissionsOptions,
+  AddEnvironmentRolePermissionOptions,
 } from './interfaces';
 import {
   deserializeEnvironmentRole,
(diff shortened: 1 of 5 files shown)


src/authorization/authorization.ts
```diff
@@ -0,0 +1,82 @@
...
+      `/authorization/roles/${slug}/permissions`,
+      { permissions },
+    );
+    return deserializeEnvironmentRole(data);
+  }
+
+  async addEnvironmentRolePermission(
+    slug: string,
+    permissionSlug: string,
+  ): Promise<EnvironmentRole> {
```
mattgd, line 77: Since these are both `string` types, it may be clearer to make this an object parameter to prevent parameter confusion. Not sure if you'd want to make changes to the other method signatures to align.

csrbarber: Yeah this is a JS style question which I'm not sure about. For POST I figured since organizationId is related to the path and not the object being created it should be a separate argument. Similar to this one, I figured since slug is related to the path and not the object, having them separate here made sense.

src/authorization/authorization.ts
```diff
@@ -0,0 +1,82 @@
...
+    const { data } = await this.workos.patch<EnvironmentRoleResponse>(
+      `/authorization/roles/${slug}`,
+      serializeUpdateEnvironmentRoleOptions(options),
+    );
+    return deserializeEnvironmentRole(data);
+  }
+
+  async setEnvironmentRolePermissions(
+    slug: string,
+    permissions: string[],
```
csrbarber: Make this an object so it's clear that the second arg applies to the action you're taking. Same as below

src/authorization/authorization.ts
csrbarber: Resolved in call - @mattgd  and Stripe checked Stripe as a reference. I'll turn permissionSlug here into an options object.

PR #1456, opened 2026-01-30
feat: Add permissions support

csrbarber:
## Description

- Adds Permission support

## Documentation

```
[X] Yes
```


src/authorization/interfaces/list-permissions-options.interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface ListPermissionsOptions {
+  before?: string;
+  after?: string;
+  limit?: number;
+  order?: 'asc' | 'desc';
+}
```
mattgd: We can extend [PaginationOptions](https://github.com/workos/workos-node/blob/main/src/common/interfaces/pagination-options.interface.ts) here.

csrbarber: Yeah good call

PR #1458, opened 2026-02-01
Add custom_attributes field to OrganizationMembership

ajworkos:
## Summary

Adds `custom_attributes` field to `OrganizationMembership` model to expose custom attributes from identity providers.

## Changes

- Added `custom_attributes` field to OrganizationMembership type/model
- Field type: Record/Map/Dictionary of string keys to any values
- Always present, defaults to empty object `{}`
- Updated fixtures and tests

## API Field Details

The field will be present in:
- REST API responses
- Webhook events
- Events API responses

JSON field name: `custom_attributes` (snake_case)

### DIFF
--- a/src/actions/actions.spec.ts
+++ b/src/actions/actions.spec.ts
@@ -155,7 +155,7 @@ describe('Actions', () => {
             slug: 'member',
           },
           status: 'active',
-          idpAttributes: {},
+          customAttributes: {},
           createdAt: '2024-10-22T17:12:50.746Z',
           updatedAt: '2024-10-22T17:12:50.746Z',
         },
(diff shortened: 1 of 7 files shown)


src/actions/actions.spec.ts
```diff
@@ -151,10 +153,12 @@ describe('Actions', () => {
           id: '01JATCNVYCHT1SZGENR4QTXKRK',
           userId: '01JATCHZVEC5EPANDPEZVM68Y9',
           organizationId: '01JATCMZJY26PQ59XT9BNT0FNN',
+          organizationName: undefined,
```
ajworkos: Tests were failing missng expectations of lastSignInAt, locale, and organizationName

It's unclear to me why these started failing in this PR, they are unrelated. Investigating

PR #1478, opened 2026-02-11
resource memberships

swaroopAkkineniWorkos:
linear: https://linear.app/workos/issue/ENT-4372/sdk-updates

I decided to break up the work for ENT-4372 into a smaller pr's that we can be easily reviewed and merge them into ENT-4372-base-authorization-branch. Then we can have one final merge that merges ENT-4372-base-authorization-branch into the main.

desc: the goal of this pr is to implement the following endpoints in the node sdk.

listResourcesForMembership() | GET /authorization/organization_memberships/{om_id}/resources
listMembershipsForResource() | GET /authorization/resources/{resource_id}/organization_memberships
listMembershipsForResourceByExternalId() | GET /authorization/organizations/{org_id}/resources/{type}/{external_id}/organization_memberships

listResourcesForMembership() ~ https://github.com/workos/workos/blob/main/packages/api/src/authorization/authorization.controller.ts#L221
listOrganizationMembershipsForResource() ~ https://github.com/workos/workos/blob/main/packages/api/src/authorization-resources/authorization-resources.controller.ts#L411
listOrganizationMembershipsForResourceByExternalId() ~ https://github.com/workos/workos/blob/main/packages/api/src/authorization-resources/authorization-resources-by-external-id.controller.ts#L135

### DIFF
--- a/src/user-management/serializers/organization-membership.serializer.ts
+++ b/src/user-management/serializers/organization-membership.serializer.ts
@@ -16,7 +16,7 @@ export const deserializeOrganizationMembership = (
   status: organizationMembership.status,
   createdAt: organizationMembership.created_at,
   updatedAt: organizationMembership.updated_at,
-  ...(organizationMembership.role && { role: organizationMembership.role }),
+  role: organizationMembership.role,
   ...(organizationMembership.roles && { roles: organizationMembership.roles }),
   customAttributes: organizationMembership.custom_attributes ?? {},
 });


src/authorization/interfaces/organization-membership-list.interface.ts
```diff
@@ -0,0 +1,64 @@
...
+export interface OrganizationMembershipListResponse {
+  object: 'list';
+  data: OrganizationMembershipResponse[];
+  list_metadata: {
+    before: string | null;
+    after: string | null;
+  };
+}
+
+export interface AuthorizationOrganizationMembership {
```
atainter: I think we already have types for organization membership because it's used with the organization memberships API. It uses the same JSON type so we should probably use the same interface.

src/authorization/interfaces/organization-membership-list.interface.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  OrganizationMembership,
+  OrganizationMembershipResponse,
+} from '../../user-management/interfaces/organization-membership.interface';
+
+export interface OrganizationMembershipList {
+  object: 'list';
+  data: OrganizationMembership[];
```
atainter: I think we'll need to create a base type because this interface includes a non-optional `role` field but it's not in the server response

https://github.com/workos/workos-node/blob/main/src/user-management/interfaces/organization-membership.interface.ts#L28

vs 

https://github.com/workos/workos/blob/main/packages/api-json-models/src/rest-api/userland-user-organization-membership.ts#L8

src/authorization/authorization.ts
```diff
@@ -400,8 +408,9 @@ export class Authorization {
     return deserializeRoleAssignment(data);
   }
 
+  // test this
   async removeRole(options: RemoveRoleOptions): Promise<void> {
-    await this.workos.delete(
+    await this.workos.deleteWithBody(
```
swaroopThereItIs-workos, line 413: changed to deleteWithBody because the delete role assignment API expects the identifying fields in the request body, and delete() doesn't send a body.

https://github.com/workos/workos/blob/1bf21e074e2cde476d83660ca41e05853fa081b8/packages/api/src/authorization/authorization-role-assignments.controller.ts#L205

src/authorization/serializers/authorization-organization-membership.serializer.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  AuthorizationOrganizationMembership,
+  AuthorizationOrganizationMembershipResponse,
+} from '../interfaces/organization-membership-list.interface';
+
+export const deserializeAuthorizationOrganizationMembership = (
```
atainter: This should be in the userland package for base organization membership

src/authorization/interfaces/organization-membership-list.interface.ts
```diff
@@ -0,0 +1,30 @@
...
+  BaseOrganizationMembership,
+  BaseOrganizationMembershipResponse,
+} from '../../user-management/interfaces/organization-membership.interface';
+
+export interface AuthorizationOrganizationMembership extends BaseOrganizationMembership {
+  organizationName?: string;
+}
+
+export interface AuthorizationOrganizationMembershipResponse extends BaseOrganizationMembershipResponse {
+  organization_name?: string;
```
atainter: I think we should remove this because authz doesn't use it

src/authorization/serializers/list-resources-for-membership-options.serializer.ts
```diff
@@ -0,0 +1,21 @@
+import {
+  ListResourcesForMembershipOptions,
+  SerializedListResourcesForMembershipOptions,
+} from '../interfaces/list-resources-for-membership-options.interface';
+
+export const serializeListResourcesForMembershipOptions = (
+  options: ListResourcesForMembershipOptions,
+): SerializedListResourcesForMembershipOptions => ({
+  permission_slug: options.permissionSlug,
+  ...(options.limit !== undefined && { limit: options.limit }),
```
atainter: Is there a common helper for the pagination params?

src/user-management/interfaces/organization-membership.interface.ts
```diff
@@ -2,30 +2,36 @@ import { RoleResponse } from '../../roles/interfaces/';
...
-  organization_name: string;
   status: OrganizationMembershipStatus;
   user_id: string;
   created_at: string;
   updated_at: string;
+  custom_attributes?: Record<string, unknown>;
+}
+
+export interface OrganizationMembershipResponse extends BaseOrganizationMembershipResponse {
+  organization_name: string;
```
atainter: this should be in the base

PR #1485, opened 2026-02-13
Add dedicated event types for organisation role deserialisation

### LINKED ISSUE #1484: `deserializeOrganizationRole` drops `organization_id` from event data, breaking Events API data sync
## Summary

The `deserializeOrganizationRole` serialiser drops the `organization_id` field from `organization_role.*` event data. This makes it impossible to determine which organisation an organisation role belongs to when consuming the Events API for data synchronisation.

All other organisation-scoped resource deserialisers (`OrganizationMembership`, `Connection`, `Session`, `OrganizationDomain`, `Invitation`, `DirectoryUser`) correctly preserve `organization_id` → `organizationId`. The `OrganizationRole` deserialiser is the sole exception.

Additionally, `deserializeRoleEvent` (used for `role.*` events) drops `id`, `name`, `description`, and `type`, which similarly prevents identifying individual roles when processing events.

## Reproduction

```ts
import { deserializeOrganizationRole } from "@workos-inc/node/lib/authorization/serializers/organization-role.serializer.js";

// Raw API response includes organization_id
// (see https://workos.com/docs/reference/events — organization_role.created)
const raw = {
  object: "organization_role",
  id: "role_01ABC",
  organization_id: "org_01XYZ",
  slug: "admin",
  name: "Admin",
  description: "Administrator role",
  resource_type_slug: "organization",
  permissions: ["users:read"],
  type: "OrganizationRole",
  created_at: "2024-01-01T00:00:00.000Z",
  updated_at: "2024-01-01T00:00:00.000Z",
};

const deserialized = deserializeOrganizationRole(raw);
console.log(deserialized.organizationId); // undefined — field is lost
```

## Root cause

[`organization-role.serializer.ts`](https://github.com/workos/workos-node/blob/main/src/authorization/serializers/organization-role.serializer.ts) constructs a new object literal and does not include `organization_id`:

```js
const deserializeOrganizationRole = (role) => ({
  object: role.object,
  id: role.id,
  name: role.name,
  slug: role.slug,
  description: role.description,
  permissions: role.permissions,
  type: "OrganizationRole",
  createdAt: role.created_at,
  updatedAt: role.updated_at,
  // organization_id is missing here
});
```

## Comparison with other deserialisers

| Resource type | Deserialiser | `organizationId` preserved? |
|---|---|---|
| `OrganizationMembership` | `deserializeOrganizationMembership` | **Yes** |
| `Connection` | `deserializeConnection` | **Yes** |
| `Session` | `deserializeSession` | **Yes** |
| `OrganizationDomain` | `deserializeOrganizationDomain` | **Yes** |
| `Invitation` | `deserializeInvitationEvent` | **Yes** |
| `DirectoryUser` | `deserializeDirectoryUser` | **Yes** |
| **`OrganizationRole`** | **`deserializeOrganizationRole`** | **No** |

## Impact

When polling the Events API (via `workos.events.listEvents()`) for data synchronisation:

1. An `organization_role.created` event arrives for a role that does not yet exist locally.
2. The deserialised `data` has no `organizationId`, so the consumer cannot determine which organisation the role belongs to.
3. The `context` field for `organization_role.*` events is `{}` (empty), so it cannot be used as a fallback.
4. The only remaining option is to scan all organisations and call `listOrganizationRoles` for each one — an O(N) operation that does not scale.

### 1. `OrganizationRole` interface and deserialiser

Add `organizationId` to the `OrganizationRole` interface and map `organization_id` in the deserialiser:

```ts
interface OrganizationRole {
  object: "role";
  id: string;
  organizationId: string; // ← add
  name: string;
  slug: string;
  description: string | null;
  permissions: string[];
  type: "OrganizationRole";
  createdAt: string;
  updatedAt: string;
}
```

```ts
const deserializeOrganizationRole = (role) => ({
  object: role.object,
  id: role.id,
  organizationId: role.organization_id, // ← add
  name: role.name,
  slug: role.slug,
  description: role.description,
  permissions: role.permissions,
  type: "OrganizationRole",
  createdAt: role.created_at,
  updatedAt: role.updated_at,
});
```

### 2. `RoleEvent` interface and deserialiser

`deserializeRoleEvent` currently returns only `slug`, `permissions`, `createdAt`, and `updatedAt`. For Events API consumers, `id`, `name`, `description`, and `type` are also needed to identify and sync individual roles. Consider aligning it with the full `Role` / `EnvironmentRole` shape.

## Environment

- `@workos-inc/node`: `8.3.1`
- Node.js / Bun runtime on Cloudflare Workers
- Use case: Events API polling for data synchronisation to PostgreSQL

smorimoto:
## Summary

- Introduces `OrganizationRoleEvent` / `OrganizationRoleEventResponse` interfaces that match the actual Events API response shape for `organization_role.*` events
- Adds `deserializeOrganizationRoleEvent` serialiser that correctly preserves `organization_id` and `resource_type_slug` from the event data
- Updates event interfaces and the event serialiser to use the new event-specific types instead of reusing REST API types

The Events API returns a different shape for organisation role data compared to the REST API (e.g. `object: "organization_role"` vs `"role"`, includes `organization_id` and `resource_type_slug`). Previously, the shared `OrganizationRole` / `OrganizationRoleResponse` types silently dropped these fields during deserialisation.

Closes #1484

## Test plan

- [x] TypeScript type check passes (`npx tsc --noEmit`)
- [x] All existing tests pass (`npm test`)
- [x] Linter and formatter pass

### DIFF
--- a/src/authorization/serializers/organization-role.serializer.ts
+++ b/src/authorization/serializers/organization-role.serializer.ts
@@ -33,15 +33,15 @@ export const deserializeOrganizationRole = (
 });
 
 export const deserializeOrganizationRoleEvent = (
-  role: OrganizationRoleEventResponse,
+  event: OrganizationRoleEventResponse,
 ): OrganizationRoleEvent => ({
-  object: role.object,
-  organizationId: role.organization_id,
-  slug: role.slug,
-  name: role.name,
-  description: role.description,
-  resourceTypeSlug: role.resource_type_slug,
-  permissions: role.permissions,
-  createdAt: role.created_at,
-  updatedAt: role.updated_at,
+  object: event.object,
+  organizationId: event.organization_id,
+  slug: event.slug,
+  name: event.name,
+  description: event.description,
+  resourceTypeSlug: event.resource_type_slug,
+  permissions: event.permissions,
+  createdAt: event.created_at,
+  updatedAt: event.updated_at,
 });


src/roles/interfaces/role.interface.ts
```diff
@@ -23,6 +23,30 @@ export interface RoleEventResponse {
   updated_at: string;
 }
 
+export interface OrganizationRoleEventResponse {
+  object: 'organization_role';
```
csrbarber, line 27:
```suggestion
  object: 'organization_role';
  id: string;
```

src/roles/interfaces/role.interface.ts
```diff
@@ -23,6 +23,30 @@ export interface RoleEventResponse {
...
+  name: string;
+  description: string | null;
+  resource_type_slug: string;
+  permissions: string[];
+  created_at: string;
+  updated_at: string;
+}
+
+export interface OrganizationRoleEvent {
+  object: 'organization_role';
```
csrbarber, line 39:
```suggestion
  object: 'organization_role';
  id: string;
```

src/authorization/serializers/organization-role.serializer.ts
```diff
@@ -26,3 +31,17 @@ export const deserializeOrganizationRole = (
...
+): OrganizationRoleEvent => ({
+  object: role.object,
+  organizationId: role.organization_id,
+  slug: role.slug,
+  name: role.name,
+  description: role.description,
+  resourceTypeSlug: role.resource_type_slug,
+  permissions: role.permissions,
+  createdAt: role.created_at,
+  updatedAt: role.updated_at,
```
csrbarber: This might be more clear as `event`. Also adding in the event id.

```suggestion
  event: OrganizationRoleEventResponse,
): OrganizationRoleEvent => ({
  object: event.object,
  organizationId: event.organization_id,
  slug: event.slug,
  name: event.name,
  description: event.description,
  resourceTypeSlug: event.resource_type_slug,
  permissions: event.permissions,
  createdAt: event.created_at,
  updatedAt: event.updated_at,
```

src/roles/interfaces/role.interface.ts
smorimoto: Really? It’s not listed here.
https://workos.com/docs/events/organization-role

src/authorization/serializers/organization-role.serializer.ts
smorimoto: That’s exactly right when it comes to naming! As for the ID, please see my other comment.

src/roles/interfaces/role.interface.ts
csrbarber: You're right! Sorry about that. I misunderstood where the event id was coming from.

src/authorization/serializers/organization-role.serializer.ts
csrbarber: You're exactly right. I've updated the suggestion to remove `id`.

smorimoto: I will rename it straight away!

smorimoto: @csrbarber Done!

PR #1502, opened 2026-02-26
feat: add `resourceTypeSlug` to permissions, environment and organization roles

csrbarber:
## Description
  - Add `resourceTypeSlug` to permission, environment role, and organization role response deserialization
  - Add optional `resourceTypeSlug` to permission and environment role creation serialization

## Documentation

```
[X] Yes
```


atainter (approved):
Should we give slug the default value of `organization`? We decided to make it optional at the API level right? Maybe we keep it optional here too

PR #1535, opened 2026-03-26
feat(oagen[1/2]): Add new endpoints to the SDK via autogeneration

gjtorikian:
## Description

This PR adds new SDK support for five new WorkOS API surfaces:

- AdminPortal: previously, this was under the folder `portal`, which is a terrible name. I `@deprecate`d those files in favor of this one with a better naming system. The SDK autogenerated _wants_ to call these `AdminPortal`.
- ~Permissions: full CRUD for authorization permissions. Previously this was confusingly tucked under `authorization`; that, too, has been marked as `@deprecate`,  considering that the SDK generator and API docs [treat them separately](https://workos.com/docs/reference/roles/permission) than the other `/authorization` endpoints.~ #1536 reverts this.
- Radar was completely missing and has been introduced
- Webhooks endpoints were missing and have been introduced
- WorkOS Connect (missing, introduced) includes `Applications` and `ApplicationClientSecrets` sub-services for managing Connect Applications and their secrets

In the cases above where the SDK autogen is choosing a name (AdminPortal, Permissions), we can, of course, override these, but I happen to agree with it. 

Note: this is the first of two PRs. 

## Why so many changed files?

Each change above was broken into its own commit. The final commit, 2b5ca3c6, introduces a set of new changes across the SDK.

Each auto-generated service class imports typed interfaces for its request/response models and enum types. These types are split into separate files by the generator (one interface per schema). The 14 dependency files break down as:

- `src/common/interfaces/`: shared enum/literal-union types (e.g., `RadarStandaloneResponseVerdict`, `WebhookEndpointJsonStatus`, `GenerateLinkIntent`) referenced
  by Radar, WebhooksEndpoints, and AdminPortal model interfaces
- `src/authorization/`: `UpdateOrganizationRole` interface and serializer, imported by the Permissions serializer because both share the same wire shape for role-permission updates


nicknisi:
@greptileai

src/workos-connect/applications.ts
```diff
@@ -0,0 +1,128 @@
...
+      '/connect/applications',
+      (() => {
+        switch ((payload as any).applicationType) {
+          case 'oauth':
+            return serializeCreateOAuthApplication(payload as any);
+          case 'm2m':
+            return serializeCreateM2MApplication(payload as any);
+          default:
+            return payload;
+        }
```
nicknisi: question: This whole switch statement casting the `payload` to `any` is making nervous. Is this hiding a larger types issue?

src/workos.ts
```diff
@@ -508,4 +518,16 @@ export class WorkOS {
...
     }
   }
+
+  readonly adminPortal = new AdminPortal(this);
+  readonly permissions = new Permissions(this);
+  readonly radar = new Radar(this);
+  readonly webhooksEndpoints = new WebhooksEndpoints(this);
+  readonly workOsConnect = new WorkOSConnect(this);
+  readonly applications = new Applications(this);
+  readonly applicationClientSecrets = new ApplicationClientSecrets(this);
```
nicknisi: question: Should these be exposed at the top-level or should it be under `workos.connect`?

src/workos.ts
```diff
@@ -508,4 +518,16 @@ export class WorkOS {
       }
     }
   }
+
+  readonly adminPortal = new AdminPortal(this);
+  readonly permissions = new Permissions(this);
+  readonly radar = new Radar(this);
+  readonly webhooksEndpoints = new WebhooksEndpoints(this);
+  readonly workOsConnect = new WorkOSConnect(this);
```
nicknisi: question: Should this just be `connect` so it's `workos.connect` instead of `workos.workOsConnect`?

If not, I'm not a fan of the `workOsConnect` casing.

src/workos.ts
nicknisi: nitpick: all these new properties should live with the other properties.

src/workos-connect/application-client-secrets.ts
```diff
@@ -0,0 +1,72 @@
...
+   * @returns {ApplicationCredentialsListItem}
+   * @throws {NotFoundException} 404
+   */
+  async list(id: string): Promise<ApplicationCredentialsListItem> {
+    const { data } =
+      await this.workos.get<ApplicationCredentialsListItemResponse>(
+        `/connect/applications/${id}/client_secrets`,
+      );
+    return deserializeApplicationCredentialsListItem(data);
+  }
```
nicknisi: question: this method is called `list` but from the looks of it only returns one item. Is this the correct name for it?

src/common/utils/fetch-and-deserialize.ts
```diff
@@ -28,3 +31,31 @@ export const fetchAndDeserialize = async <T, U>(
...
+      endpoint,
+      deserializeFn,
+      options,
+    ),
+    (params) =>
+      fetchAndDeserialize<TResponse, TModel>(
+        workos,
+        endpoint,
+        deserializeFn,
+        params,
```
nicknisi: `AutoPaginatable` only hands this callback `{ limit, after }` on each page, so any extended filter options (like `organizationId`) get dropped after page 1. Merging `options` in keeps the filters on every page, and putting `params` last lets the pagination cursor still override `after`/`limit` as pages advance.

```suggestion
        { ...options, ...params },
```

src/workos-connect/applications.ts
```diff
@@ -0,0 +1,128 @@
...
+      ConnectApplicationResponse,
+      ConnectApplication,
+      ApplicationsListOptions
+    >(
+      this.workos,
+      '/connect/applications',
+      deserializeConnectApplication,
+      options,
+    );
+  }
```
nicknisi: question: is this going to pass in the deserialized options? Like `organizationId` instead of `organization_id`?

src/portal/portal.ts
```diff
@@ -1,6 +1,7 @@
 import { WorkOS } from '../workos';
 import { GeneratePortalLinkIntent } from './interfaces/generate-portal-link-intent.interface';
 
+/** @deprecated Use `AdminPortal` instead. */
```
nicknisi, line 4: issue: this was added here but the deprecation message doesn't pass through to the `WorkOS` property (`workos.portal.generateLink`, for example). The `@deprecated` directive should be added to the property in `wokros.ts`, too.

src/authorization/authorization.ts
```diff
@@ -240,6 +241,7 @@ export class Authorization {
     return deserializePermission(data);
   }
 
+  /** @deprecated Use `workos.permissions.list()` instead. */
   async listPermissions(
     options?: ListPermissionsOptions,
   ): Promise<PermissionList> {
```
nicknisi, line 253: issue: This returns `PermissionList` which is different from what `workos.permissions.list` will return (`AutoPaginatable`). This is technically a breaking change.

src/workos-connect/applications.ts
```diff
@@ -0,0 +1,128 @@
...
+import type { UpdateOAuthApplication } from './interfaces/update-oauth-application.interface';
+import { deserializeConnectApplication } from './serializers/connect-application.serializer';
+import { serializeCreateOAuthApplication } from './serializers/create-oauth-application.serializer';
+import { serializeCreateM2MApplication } from './serializers/create-m2m-application.serializer';
+import { serializeUpdateOAuthApplication } from './serializers/update-oauth-application.serializer';
+
+export interface ApplicationsListOptions extends PaginationOptions {
+  /** Filter Connect Applications by organization ID. */
+  organizationId?: string;
+}
```
nicknisi: question: should this be externally exported? I don't think it's exposed anywhere.

src/workos.ts
```diff
@@ -1,3 +1,5 @@
+// This file is auto-generated by oagen. Do not edit.
```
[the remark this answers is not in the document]
gjtorikian: See comment [here](https://github.com/workos/workos-node/pull/1536#issuecomment-4237282431)

src/workos-connect/applications.ts
gjtorikian: I fixed this now that you and [Greptile](https://github.com/workos/workos-node/pull/1535#discussion_r3074705257) commented on it. 😆  The emitter will no longer try to "fake" its way through discriminated unions like this; it will error by `default`.

src/workos.ts
gjtorikian: sounds good to me, changed

src/workos.ts
gjtorikian: switched to `connect` (per above)

src/workos.ts
gjtorikian: good call, the emitter blindly pushes them down here. fixed.

src/workos-connect/application-client-secrets.ts
gjtorikian: yep, [Greptile flagged](https://github.com/workos/workos-node/pull/1535#discussion_r3074705330) this too. It turns out that this is a non-paginated array in the spec, so the emitter stumbled over it. fixed.

src/authorization/authorization.ts
gjtorikian: Couple of things:

- I actually removed `workos.permissions` in https://github.com/workos/workos-node/pull/1536#issuecomment-4237282431, but didn't cherry-pick the change back into this branch
- More broadly:`Authorization.listPermissions` was still exported with the same signature and return type, so I'm not sure how this is breaking. The new `workos.permissions.list` (which, again, is no longer introduced) would have worked around this by providing `AutoPaginatable<Permission>` in an alternate method.

src/workos-connect/applications.ts
gjtorikian: Addressed in commit [e3247724](https://github.com/workos/workos-node/pull/1535/commits/e3247724a97fb5c75ba30845e665f6412d2ecc86)

gjtorikian:
I'm going to break this branch up into three (to support the new Radar, Webhooks endpoints, and Connect additions) and point them at the upcoming v9 release; I dropped the soft deprecation and rolled the `adminPortal` rename into https://github.com/workos/workos-node/pull/1562.

PR #1543, opened 2026-04-06
fix(events): handle domain verification failed

noseworthy:
## Description
The domain verfication failed events have a slightly different schema to other organization domain events. This was causing the data to be parsed incorrectly in the list events api resulting in empty `data` properties.

I've added the required interfaces, serializers, and deserializers to capture the difference and parse the events successfully. I've also added some unit tests for the serialization.

## Documentation

```
[ ] Yes
```


gjtorikian:
looks great, thank you!

gjtorikian (approved):

PR #1563, opened 2026-04-20
fix: Export `ConflictException` and add `code`

davidcornu:
## Description

Some operations like `workos.authorization.deleteOrganizationRole` will `throw` a `ConflictException` that needs to be handled appropriately (e.g. to customize the error message before showing it to users).

However, `ConflictException` isn't current exported so it can't be used to perform type-safe error handling (e.g. `error instanceof ConflictException`). It also doesn't pass along the `code` property which is needed to differentiate error types (e.g. https://workos.com/docs/reference/roles/custom-role#delete-a-custom-role).

## Documentation

```
[ ] Yes
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the default wording for conflict errors to fix a typo in the message.
  * Conflict errors may now include an optional error code in error payloads.

* **Chores**
  * Made the conflict error type available via the public exception exports.

### DIFF
--- a/src/common/exceptions/conflict.exception.ts
+++ b/src/common/exceptions/conflict.exception.ts
@@ -4,15 +4,18 @@ export class ConflictException extends Error implements RequestException {
   readonly status = 409;
   readonly name = 'ConflictException';
   readonly requestID: string;
+  readonly code?: string;
 
   constructor({
     error,
     message,
     requestID,
+    code,
   }: {
     error?: string;
     message?: string;
     requestID: string;
+    code?: string;
   }) {
     super();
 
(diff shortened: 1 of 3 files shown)


src/common/exceptions/conflict.exception.ts
```diff
@@ -25,5 +28,7 @@ export class ConflictException extends Error implements RequestException {
     } else {
       this.message = `An conflict has occurred on the server.`;
```
davidcornu: Fixing this typo would probably be a breaking change 🙈

gjtorikian: we're actually on the verge of making a breaking v9 change so you're in luck! (I changed it)

PR #1573, opened 2026-04-24
fix: Correctly handle all `isAuthenticationErrorData` cases

davidcornu:
## Description

https://github.com/workos/workos-node/pull/1561 added a new `AuthenticationException` type to handle the following cases: https://github.com/workos/workos-node/blob/96cf545f973e3e130abdbc9ea09a39b4cced7af3/src/common/exceptions/authentication.exception.ts#L7-L13

However, because the error handling logic first checks whether the payload has an `error` or `error_description` field https://github.com/workos/workos-node/blob/96cf545f973e3e130abdbc9ea09a39b4cced7af3/src/workos.ts#L486-L487
`sso_required` is being treated as an `OauthException` instead of a `AuthenticationException` (see https://workos.com/docs/reference/authkit/authentication-errors#sso-required-error).

## Documentation

```
[ ] Yes
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP error classification to prioritize and reliably detect authentication failures (avoiding misclassification as OAuth or generic errors).
  * Authentication errors now surface consistent error codes and clearer messages sourced from multiple response fields.
* **Tests**
  * Added tests to verify authentication error detection and that the thrown exception preserves status, raw response details, code, and message.

### DIFF
--- a/src/common/exceptions/authentication.exception.ts
+++ b/src/common/exceptions/authentication.exception.ts
@@ -32,9 +32,18 @@ const AUTHENTICATION_ERROR_CODES: ReadonlySet<string> = new Set<string>([
 export function isAuthenticationErrorData(
   data: WorkOSErrorData,
 ): data is AuthenticationErrorData {
-  return (
-    typeof data.code === 'string' && AUTHENTICATION_ERROR_CODES.has(data.code)
-  );
+  let discriminant: string | undefined;
+
+  if (typeof data.code === 'string') {
+    discriminant = data.code;
+  } else if (typeof data.error === 'string') {
+    // Some errors like `sso_required` use an `error` field instead of `code`
+    discriminant = data.error;
+  }
+
+  if (!discriminant) return false;
+
+  return AUTHENTICATION_ERROR_CODES.has(discriminant);
 }
 
 export class AuthenticationException extends GenericServerException {
--- a/src/common/exceptions/authentication.exception.ts
+++ b/src/common/exceptions/authentication.exception.ts
@@ -12,13 +12,23 @@ export type AuthenticationErrorCode =
   | 'mfa_verification'
   | 'sso_required';
 
-export interface AuthenticationErrorData extends WorkOSErrorData {
-  code: AuthenticationErrorCode;
+interface BaseAuthenticationErrorData extends WorkOSErrorData {
+  error?: string;
+  error_description?: string;
   pending_authentication_token?: string;
   user?: UserResponse;
   organizations?: Array<{ id: string; name: string }>;
   connection_ids?: string[];
 }
+
+export type AuthenticationErrorData =
+  | (BaseAuthenticationErrorData & {
+      code: AuthenticationErrorCode;
+    })
+  | (BaseAuthenticationErrorData & {
+      code?: AuthenticationErrorCode;
+      error: AuthenticationErrorCode;
+    });
 
 const AUTHENTICATION_ERROR_CODES: ReadonlySet<string> = new Set<string>([
   'email_verification_required',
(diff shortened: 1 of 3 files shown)


gjtorikian:
## Review Notes

Nice fix — the reordering in `workos.ts` is correct and the motivation is clear. Two things worth addressing before merging:

### 1. Type guard unsoundness on `AuthenticationErrorData.code`

`isAuthenticationErrorData` now returns `true` when the discriminant comes from `data.error` (not `data.code`), but the `AuthenticationErrorData` interface declares `code: AuthenticationErrorCode` as **non-optional**. After the type guard narrows, TypeScript consumers will believe `rawData.code` is always an `AuthenticationErrorCode` string — but `sso_required` responses definitively have no `code` field (confirmed against the API source and OpenAPI spec). The actual response shape is:

```json
{
  "error": "sso_required",
  "error_description": "User must authenticate using one of the matching connections.",
  "email": "user@example.com",
  "connection_ids": ["conn_01FVYZ..."]
}
```

Any consumer doing something like:

```ts
if (e instanceof AuthenticationException) {
  switch (e.rawData.code) {  // TypeScript says AuthenticationErrorCode, runtime says undefined
    case 'sso_required': // never matches
  }
}
```

would silently get `undefined` with no compiler warning.

**Suggestion:** Make `code` optional in `AuthenticationErrorData` and add `error?` as a field so the types are honest about what's present.

### 2. Regression test (echoing greptile-apps bot)

+1 to the bot's suggestion — a test with a payload matching the actual API response shape would lock in this fix against future refactors:

```ts
{ error: 'sso_required', error_description: 'User must authenticate using one of the matching connections.', email: 'user@example.com', connection_ids: ['conn_123'] }
```

asserting `AuthenticationException` is thrown (not `OauthException`).

davidcornu:
Thanks for the review @gjtorikian. For context - I have no attachment to my current implementation, just figured I'd put together a quick fix as I've unfortunately become all-too-familiar with this part of the SDK over the last few months and the upgrade to v9 caused production errors for us (we got a `OauthException` where we were expecting a `AuthenticationException`). If you want to throw this away and implement your own fix don't hesitate to.

----

> `isAuthenticationErrorData` now returns `true` when the discriminant comes from `data.error` (not `data.code`), but the `AuthenticationErrorData` interface declares `code: AuthenticationErrorCode` as **non-optional**. After the type guard narrows, TypeScript consumers will believe `rawData.code` is always an `AuthenticationErrorCode` string — but `sso_required` responses definitively have no `code` field (confirmed against the API source and OpenAPI spec).

Good point. Thanks for checking the source!

> **Suggestion:** Make `code` optional in `AuthenticationErrorData` and add `error?` as a field so the types are honest about what's present.

This doesn't make for a great experience for downstream users as they now have to care about whether `AuthenticationErrorData` has a `code` or an `error` in order to correctly handle it. I would either

- add a field on `AuthenticationException` that always returns a `AuthenticationErrorCode` (maybe `code`)
    - also add `error?: string` as you suggested
        - if we want to be more specific this could be `"sso_required"` instead of `string`
    - make the `code` field `code?: string` or `Exclude<AuthenticationErrorCode, "sso_required">` to be more correct
- keep `sso_required` as an `OauthException` (although that doesn't make much sense to me).

thoughts?

> +1 to the bot's suggestion — a test with a payload matching the actual API response shape would lock in this fix against future refactors:
> [...]
> asserting `AuthenticationException` is thrown (not `OauthException`).

Sure thing. Surprised this wasn't flagged on the original PR.

gjtorikian:
Yeah, I do agree that users shouldn't have to check both `rawData.code` and `rawData.error`. It mostly stinks because the API alternates between `error/error_description` and `code/message`.

As I took notes across the PR and the spec I realized it would probably just be easier for me to push the changes here, so. Consumers can just use `e.code` going forward.

I'm sort of speed running this (it's late on a Friday!), but the changes I made are, at least, technically accurate. Let me know if they solve your use case and I can push a release.

davidcornu:
> I'm sort of speed running this (it's late on a Friday!), but the changes I made are, at least, technically accurate. Let me know if they solve your use case and I can push a release.

No rush. Solved my issue by reverting some of our v9 changes and adding a code comment that points to this PR 😬.

src/common/exceptions/authentication.exception.ts
```diff
@@ -12,14 +12,24 @@ export type AuthenticationErrorCode =
...
 
+export type AuthenticationErrorData =
+  | (BaseAuthenticationErrorData & {
+      code: AuthenticationErrorCode;
+    })
+  | (BaseAuthenticationErrorData & {
+      code?: AuthenticationErrorCode;
+      error: AuthenticationErrorCode;
+    });
+
```
davidcornu, line 30: Returning `AuthenticationErrorCode` isn't actually correct (and might lead to downstream users implementing a lot more cases than actually exist).

If `sso_required` is the only case where `error` and `error_description` are present then maybe this can be a bit more specific?

```ts
interface BaseAuthenticationErrorData extends WorkOSErrorData {
  pending_authentication_token?: string;
  user?: UserResponse;
  organizations?: Array<{ id: string; name: string }>;
  connection_ids?: string[];
}

export type AuthenticationErrorData =
  | (BaseAuthenticationErrorData & {
      code: Exclude<AuthenticationErrorCode, "sso_required">;
    })
  | (BaseAuthenticationErrorData & {
      error: "sso_required";
      error_description: string;
    });

```

src/common/exceptions/authentication.exception.ts
```diff
@@ -29,24 +39,60 @@ const AUTHENTICATION_ERROR_CODES: ReadonlySet<string> = new Set<string>([
...
 
+export function isAuthenticationErrorData(
+  data: WorkOSErrorData,
+): data is AuthenticationErrorData {
+  return getAuthenticationErrorCode(data) !== undefined;
+}
+
 export class AuthenticationException extends GenericServerException {
   readonly name = 'AuthenticationException';
+  override readonly code: AuthenticationErrorCode;
```
davidcornu, line 77: This is definitely a better user experience.

src/common/exceptions/authentication.exception.ts
gjtorikian: I like it 👍

davidcornu:
Thanks @gjtorikian ❤️

PR #1578, opened 2026-05-01
fix(webhooks): accept raw request bytes for signature verification

nicknisi:
## Summary
- Widens `constructEvent` and `verifyHeader` payload type to accept `string | Uint8Array | ArrayBuffer` in addition to parsed objects
- Raw-bytes paths HMAC the exact bytes passed in; object path preserves existing `JSON.stringify` behavior for back-compat
- `constructEvent` parses after verification on raw-bytes paths
- Shared `decodePayloadToString` helper ensures signing and parsing always use identical decoding
- Uses realm-agnostic binary detection (`ArrayBuffer.isView` / `Object.prototype.toString.call`) instead of `instanceof` checks, so payloads from Workers, iframes, or VM contexts are handled correctly

## Test plan
- [x] Existing webhook tests pass (legacy object path)
- [x] New tests for raw string, Buffer, Uint8Array, and ArrayBuffer payloads
- [x] Divergence regression tests (whitespace, unicode escapes, key reordering, BOM injection)

## Documentation

```
[x] Yes
```

WorkOS Docs webhook snippets need updating to pass raw request bodies instead of `JSON.parse`-ing before `constructEvent` (separate docs PR).

Link to Devin session: https://app.devin.ai/sessions/1170d117f26f431bbc866077728682f6
Requested by: @nicknisi

### DIFF
--- a/src/webhooks/webhooks.ts
+++ b/src/webhooks/webhooks.ts
@@ -11,7 +11,11 @@ import {
 // Parse only after verification succeeds — a malformed body never reaches
 // JSON.parse on an unauthenticated request.
 function parseVerifiedPayload(payload: WebhookPayload): EventResponse {
-  if (typeof payload === 'object' && !(payload instanceof Uint8Array) && !(payload instanceof ArrayBuffer)) {
+  if (
+    typeof payload === 'object' &&
+    !(payload instanceof Uint8Array) &&
+    !(payload instanceof ArrayBuffer)
+  ) {
     return payload as unknown as EventResponse;
   }
   return JSON.parse(decodePayloadToString(payload)) as EventResponse;
--- a/src/webhooks/webhooks.ts
+++ b/src/webhooks/webhooks.ts
@@ -3,28 +3,18 @@ import { deserializeEvent } from '../common/serializers';
 import { Event, EventResponse } from '../common/interfaces';
 import { SignatureProvider } from '../common/crypto/signature-provider';
 import { CryptoProvider } from '../common/crypto/crypto-provider';
+import {
+  type WebhookPayload,
+  decodePayloadToString,
+} from '../common/crypto/decode-payload';
 
 // Parse only after verification succeeds — a malformed body never reaches
 // JSON.parse on an unauthenticated request.
-function parseVerifiedPayload(
-  payload: string | Uint8Array | ArrayBuffer | Record<string, unknown>,
-): EventResponse {
-  if (typeof payload === 'string') {
-    return JSON.parse(payload) as EventResponse;
+function parseVerifiedPayload(payload: WebhookPayload): EventResponse {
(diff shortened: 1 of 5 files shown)


src/webhooks/webhooks.ts
```diff
@@ -31,15 +52,19 @@ export class Webhooks {
     secret,
     tolerance = 180000,
   }: {
-    payload: Record<string, unknown>;
+    // Prefer raw request bytes (string / Uint8Array / Buffer / ArrayBuffer) so
+    // the HMAC is computed over the exact bytes WorkOS signed. Parsed objects
+    // are still accepted for back-compat but fall through a JSON.stringify
+    // round-trip that can disagree with the on-the-wire bytes.
+    payload: string | Uint8Array | ArrayBuffer | Record<string, unknown>;
```
gjtorikian: I believe this is missing ` | unknown`, to match the above?

identity-wael:
Independent reproduction from production (us) — confirming this is the right fix.

Hit this today while wiring `@workos-inc/node@7.46.0`'s `webhooks.constructEvent({ payload: rawBody, sigHeader, secret })` against a freshly-created webhook endpoint. Symptom: every delivery rejected with `Signature hash does not match the expected signature hash for payload`.

Diagnostic confirmed the SDK is HMAC'ing different bytes than WorkOS signs:

- Manual `createHmac('sha256', secret).update(`${t}.${rawBody}`).digest('hex')` → matches the `v1=` value in the `WorkOS-Signature` header byte-for-byte.
- SDK call with the same `rawBody` string → mismatched HMAC.

Root cause is the unconditional `payload = JSON.stringify(payload)` at `signature-provider.js` line 47, which double-encodes when `payload` is already the raw request body string. The SDK's own spec masks this because it imports a JS object via `require('./fixtures/...')` and then computes the test signature with `JSON.stringify(payload)` — so the test fixture round-trips through the same broken logic.

Quick repro (any Node):
```js
const { createHmac } = require('crypto');
const secret = 's', t = '1777940637677';
const rawBody = '{"id":"event_01","data":{"name":"foo"}}';

const correct = createHmac('sha256', secret).update(`${t}.${rawBody}`).digest('hex');
const sdkLike = createHmac('sha256', secret).update(`${t}.${JSON.stringify(rawBody)}`).digest('hex');
console.log({ correct, sdkLike, match: correct === sdkLike }); // match: false
```

Working around it on our end with a manual verifier (parse `t=`, `v1=`, HMAC over `${t}.${rawBody}`, `crypto.timingSafeEqual`). Will switch back to the SDK once #1578 ships in a release. Happy to test against a pre-release if useful.

Also worth noting: the `WorkOS-Signature` header's `t=` value carries **millisecond** epoch (13 digits). The current `parseInt(timestamp, 10) < Date.now() - tolerance` check at `verifyHeader` happens to work because both sides are ms — but `tolerance = 180000` (ms) does the right thing. Just flagging in case it surfaces during the #1578 review since #1579 went after the same area.

gjtorikian (approved):

PR #1596, opened 2026-05-20
feat: Add Radar to Node SDK

gjtorikian:
## Summary
- Adds the Radar module to the Node SDK, exposing standalone Radar assess, update-attempt, update-list, and delete-list-entry operations on `workos.radar`.
- Introduces interfaces, serializers, and JSON fixtures for Radar request/response payloads (verdicts, controls, actions, auth methods, blocklist types, etc.).
- Registers Radar in `src/index.ts`, `src/workos.ts`, and `.oagen-manifest.json` so generated artifacts and barrel exports stay in sync.
- Adds `radar.spec.ts` covering the new client methods.

## Test plan
- [x] `npm test -- radar.spec.ts` passes
- [x] Full test suite passes
- [x] `workos.radar` is reachable from a consumer and types resolve correctly

## Summary by CodeRabbit

* **New Features**
  * Added Radar API client to the SDK for creating and updating security assessment attempts.
  * Added allowlist/blocklist management endpoints (add/remove entries).
  * SDK now exposes the Radar feature on the main client.

* **Bug Fixes / Improvements**
  * Added webhook events for Pipes connected account lifecycle: connected, disconnected, reauthorization needed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/workos/workos-node/pull/1596?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/radar/interfaces/radar-action.interface.ts
```diff
@@ -0,0 +1,8 @@
+// This file is auto-generated by oagen. Do not edit.
+
+export const RadarAction = {
+  Block: 'block',
```
blairworkos: We can also issue a `Challenge: 'challenge'` action here

blairworkos: oh I see - this is just for lists, where we only support block and allow. Could we update the const naming to be something like "RadarListAction"?

src/radar/interfaces/radar-standalone-assess-request-action.interface.ts
```diff
@@ -0,0 +1,11 @@
+// This file is auto-generated by oagen. Do not edit.
+
+export const RadarStandaloneAssessRequestAction = {
```
blairworkos, line 3: Could we narrow this to just be "sign-up" and "sign-in"? I'm also looking at the public docs, and we support like a lot of different variations, but I think that's just iterations. Wonder if we can only show these two options publicly to try and get users to send the right one.

src/radar/interfaces/radar-standalone-response-control.interface.ts
```diff
@@ -0,0 +1,17 @@
+// This file is auto-generated by oagen. Do not edit.
+
+export const RadarStandaloneResponseControl = {
+  BotDetection: 'bot_detection',
+  BruteForceAttack: 'brute_force_attack',
+  CredentialStuffing: 'credential_stuffing',
+  DomainSignUpRateLimit: 'domain_sign_up_rate_limit',
+  IpSignUpRateLimit: 'ip_sign_up_rate_limit',
```
blairworkos: hmmm this ip_sign_up_rate_limit should have been deprecated - I think that's something we'll need to do in the API.

blairworkos (approved):

PR #1620, opened 2026-06-16
feat!: generate OrganizationDomains off the OpenAPI spec

gjtorikian:
## Description

This PR generates the OrganizationDomain category in the Node SDK from the OpenAPI spec.

  Three of the four methods switched from a positional id: string to an options object:

  ```
 ┌──────────────────────────┬────────────────────────────────────────────┬────────────────────────────────────────────┐
  │          Method          │                   Before                   │                   After                    │
  ├──────────────────────────┼────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ getOrganizationDomain    │ (id: string)                               │ (options: { id: string })                  │
  ├──────────────────────────┼────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ verifyOrganizationDomain │ (id: string)                               │ (options: { id: string })                  │
  ├──────────────────────────┼────────────────────────────────────────────┼────────────────────────────────────────────┤
  │ deleteOrganizationDomain │ (id: string)                               │ (options: { id: string })                  │
  ├──────────────────────────┼────────────────────────────────────────────┼────────────────────────────────────────────┤
  ```
  
  As well, the following output responses changed:
  
  - `createdAt` / `updatedAt`: string → Date. The serializer now does new Date(response.created_at). Anyone treating these as ISO strings (e.g. string ops, direct JSON passthrough) will break.
  - `state` and `verificationStrategy` are now optional (state?, verificationStrategy?) — they can be undefined where before they were guaranteed present.
  
  Because of this, the PR should be treated as a breaking change.


src/organization-domains/organization-domains.ts
```diff
@@ -1,88 +1,102 @@
...
+import type { WorkOS } from '../workos';
+import type { CreateOrganizationDomainOptions } from './interfaces/create-organization-domain-options.interface';
+import type { GetOrganizationDomainOptions } from './interfaces/get-organization-domain-options.interface';
+import type { DeleteOrganizationDomainOptions } from './interfaces/delete-organization-domain-options.interface';
+import type { VerifyOrganizationDomainOptions } from './interfaces/verify-organization-domain-options.interface';
+import type {
   OrganizationDomain,
   OrganizationDomainResponse,
-} from './interfaces';
-import { serializeCreateOrganizationDomainOptions } from './serializers/create-organization-domain-options.serializer';
```
kevinferri, line 7: is `serializeCreateOrganizationDomainOptions` dead code now?

src/organization-domains/organization-domains.spec.ts
```diff
@@ -1,110 +1,103 @@
...
-import { fetchOnce, fetchURL, fetchBody } from '../common/utils/test-utils';
+import {
+  fetchOnce,
+  fetchURL,
+  fetchMethod,
+  fetchBody,
+} from '../common/utils/test-utils';
 import { WorkOS } from '../workos';
-import getOrganizationDomainPending from './fixtures/get-organization-domain-pending.json';
-import getOrganizationDomainVerified from './fixtures/get-organization-domain-verified.json';
```
kevinferri, line 5: are getOrganizationDomainPending and getOrganizationDomainVerified dead code now?

src/organization-domains/organization-domains.ts
```diff
@@ -1,88 +1,102 @@
-import { WorkOS } from '../workos';
-import {
-  CreateOrganizationDomainOptions,
+// This file is auto-generated by oagen. Do not edit.
+
+import type { WorkOS } from '../workos';
+import type { CreateOrganizationDomainOptions } from './interfaces/create-organization-domain-options.interface';
```
kevinferri, line 4: serializeCreateOrganizationDomain takes the generated CreateOrganizationDomain, but this method types its input as the hand-owned CreateOrganizationDomainOptions which has the same shape ({ domain, organizationId }), so we now carry two interfaces for one input type (and SerializedCreateOrganizationDomainOptions is now fully orphaned). Can the client use CreateOrganizationDomain directly so we can delete create-organization-domain-options.interface.ts?

src/organization-domains/serializers.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+import createOrganizationDomainFixture from './fixtures/create-organization-domain.json';
+import organizationDomainStandAloneFixture from './fixtures/organization-domain-stand-alone.json';
+import organizationDomainFixture from './fixtures/organization-domain.json';
+
+describe('CreateOrganizationDomainSerializer', () => {
+  it('serializes correctly', () => {
+    const fixture =
+      createOrganizationDomainFixture as CreateOrganizationDomainResponse;
+    const serialized = serializeCreateOrganizationDomain(fixture as any);
+    expect(serialized).toBeDefined();
```
kevinferri, line 15: Should we assert the actual output here? toBeDefined() would pass even if the mapping were off. The fixture is snake_case (organization_id) but the serializer reads model.organizationId, so serialized.organization_id ends up undefined.

kevinferri (approved):

PR #1627, opened 2026-06-17
feat!: Port AuditLogs to oagen

gjtorikian:
## Description

This PR generates the AuditLogs category in the Node SDK from the OpenAPI spec.

Almost every existing method changed its input signature, its output, or both. Callers cannot expect the same input/output format.

  ```
┌──────────────┬─────────────────────────────────────────────────┬────────────────────────────────────────────────────────────┐
  │    Method    │                     Before                      │                           After                            │
  ├──────────────┼─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ createEvent  │ createEvent(organization: string, event,        │ createEvent(options: CreateEventOptions) →                 │
  │              │ options?) → Promise<void>                       │ Promise<AuditLogEventCreateResponse>                       │
  ├──────────────┼─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ createSchema │ createSchema(schema: {action, ...}, options?)   │ createSchema(options: {actionName, ...})                   │
  ├──────────────┼─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ listSchemas  │ listSchemas(action: string, options?)           │ renamed → listActionSchemas(options: {actionName, ...})    │
  ├──────────────┼─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ getExport    │ getExport(auditLogExportId: string)             │ getExport(options: {auditLogExportId})                     │
  ├──────────────┼─────────────────────────────────────────────────┼────────────────────────────────────────────────────────────┤
  │ createExport │ createExport(options: AuditLogExportOptions)    │ same signature                                             │
  └──────────────┴─────────────────────────────────────────────────┴────────────────────────────────────────────────────────────┘
  ```
  
  As well:
  
  - `AuditLogExport.createdAt` / `updatedAt` change: `string → Date`
  - `AuditLogSchema.createdAt` also changed `string → Date`

The following new methods were added:

  - `getOrganizationAuditLogsRetention(options)`
  - `updateOrganizationAuditLogsRetention(options)`
  - `listActions(options?)`
  
  Because of these changes, this is a major breaking change to the SDK.


src/audit-logs/interfaces/audit-log-export-options.interface.ts
```diff
@@ -3,17 +3,7 @@ export interface AuditLogExportOptions {
...
-}
-
-export interface SerializedAuditLogExportOptions {
-  actions?: string[];
-  actor_names?: string[];
-  actor_ids?: string[];
-  organization_id: string;
-  range_end: string;
-  range_start: string;
+  rangeEnd: string;
```
swaroopAkkineniWorkos, line 6: is there a reason to make this string in lieu of date?

src/audit-logs/audit-logs.ts
```diff
@@ -1,36 +1,197 @@
...
+   *
+   * Get a list of all schemas for the Audit Logs action identified by `:name`.
+   * @param options - Pagination and filter options.
+   * @param options.actionName - The name of the Audit Log action.
+   * @example "user.logged_in"
+   * @returns {Promise<AutoPaginatable<AuditLogSchema, Omit<ListActionSchemasOptions, "actionName">>>}
+   * @throws {NotFoundException} 404
+   * @throws {UnprocessableEntityException} 422
+   */
+  async listActionSchemas(
```
swaroopAkkineniWorkos, line 142: seems like a pretty big breaking change to change the name from listSchemas -> ListActionSchemas

src/audit-logs/audit-logs.ts
```diff
@@ -1,36 +1,197 @@
-import { PaginationOptions } from '../common/interfaces';
-import { fetchAndDeserialize } from '../common/utils/fetch-and-deserialize';
+// This file is auto-generated by oagen. Do not edit.
+
+import type { WorkOS } from '../workos';
+import type { PaginationOptions } from '../common/interfaces/pagination-options.interface';
 import { AutoPaginatable } from '../common/utils/pagination';
-import { WorkOS } from '../workos';
-import {
-  CreateAuditLogEventOptions,
```
swaroopAkkineniWorkos, line 6: the new export structure will break some implementations

swaroopAkkineniWorkos (approved):
a bit worried about the breaeking changes

src/audit-logs/interfaces/audit-log-export-options.interface.ts
gjtorikian: No; this should be corrected in the OpenAPI spec.

src/audit-logs/audit-logs.ts
gjtorikian: I can rename this 👍

ConderShou:
Commenting on this thread as a follow-up to this issue: https://github.com/workos/workos-node/issues/1704

The two retention SDK methods added here are currently the only path to `GET`/`PUT /organizations/{id}/audit_logs_retention` from Node. If we're worried about the breaking changes, would you consider splitting these three methods (`getOrganizationAuditLogsRetention`, `updateOrganizationAuditLogsRetention`, `listActions`) into a separate PR? I imagine these shouldn't need anything major.

gjtorikian:
Yes, I can pull those out--will do so.

PR #1630, opened 2026-06-17
refactor(feature-flags): replace eventemitter3 with internal typed emitter

nicknisi:
## What

Removes the `eventemitter3` runtime dependency and replaces it with a small, first-party typed `EventEmitter` (`src/feature-flags/event-emitter.ts`). `FeatureFlagsRuntimeClient` now extends this internal emitter and declares a typed event map (`change` / `error` / `failed`), so event names and payloads are compile-checked.

After this change the SDK has **zero runtime dependencies** (`dependencies: {}`).

## Why

`eventemitter3` lived in `dependencies` and was **not** on tsdown's `onlyAllowBundle` whitelist, so every build artifact — including the `workerd` / `edge-light` / `convex` entries — emitted a bare `import { EventEmitter } from 'eventemitter3'`. That external specifier had to be resolved by the consumer's bundler/runtime, which is what made it leak into client/edge bundles. As inlined first-party code there is no external module to resolve, no `exports`-map condition matching, and no dual-package hazard — the emitter is compiled directly into each artifact.

(`eventemitter3` is edge-safe at runtime; the problem was packaging, not the library.)

## Behavior notes

- **Unhandled `error` throws.** The internal emitter throws on an emitted `error` with no listener (matching Node's `EventEmitter`), replacing the old `override emit` shim on the client. Poll failures are never silently dropped.
- **Public surface is a superset of eventemitter3's.** In addition to `on`/`once`/`off`/`emit`/`listenerCount`/`removeAllListeners`, the emitter keeps `addListener`/`removeListener`/`listeners`/`eventNames` aliases so swapping the base class is non-breaking for consumers that used them.
- **Dispatch safety:** listeners are snapshotted at `emit` time, so adding/removing listeners (including `once` auto-removal) mid-dispatch is well-defined and covered by tests.

## Testing

- New `event-emitter.spec.ts` — on/once/off, registration order, dispatch-during-dispatch safety, unhandled-`error` throwing, and the eventemitter3 aliases.
- `runtime-client.spec.ts` — added a guard test that `emit('error')` with no listener throws.
- `npx tsc --noEmit`, prettier, and eslint all clean; full feature-flags suite green.

## Follow-up (not in this PR)

Pre-existing: in `poll()`, `emit('error')` with no listener attached throws from inside the catch block, which skips `scheduleNextPoll()` and can halt polling (and surface as an unhandled rejection, since `poll` runs in a timer). This behavior predates this PR — it carried over from the old `override emit` — but is worth a separate look at whether a background poller should be able to be wedged by a missing observer.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,9 +8,6 @@
       "name": "@workos-inc/node",
       "version": "10.2.1",
       "license": "MIT",
-      "dependencies": {
-        "eventemitter3": "^5.0.4"
-      },
       "devDependencies": {
         "@arethetypeswrong/cli": "^0.18.2",
         "@babel/plugin-transform-modules-commonjs": "^7.26.3",
@@ -6307,12 +6304,6 @@
       "engines": {
         "node": ">=0.10.0"
       }
-    },
-    "node_modules/eventemitter3": {
-      "version": "5.0.4",
-      "resolved": "https://registry.npmjs.org/eventemitter3/-/eventemitter3-5.0.4.tgz",
-      "integrity": "sha512-mlsTRyGaPBjPedk6Bvw+aqbsXDtoAyAzm5MO7JgU+yVRyMQ5O8bD4Kcci7BS85f93veegeCPkL8R4GLClnjLFw==",
-      "license": "MIT"
     },
     "node_modules/execa": {
       "version": "5.1.1",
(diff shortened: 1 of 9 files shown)


gjtorikian (approved):
lgtm!

PR #1642, opened 2026-06-24
feat: Add agent registration read and credential validation methods

m0tzy:
## Summary

Adds two SDK methods for the new API-key-authenticated agent-auth endpoints introduced in workos/workos#63723, surfaced under `workos.agents`:

- **`workos.agents.getRegistration(id)`** → `GET /agents/registrations/:id`. Reads a single agent registration scoped to the API key's environment, returning its derived `status`/`kind`, agent identity, organization, and claim state.
- **`workos.agents.validateCredential({ type, credential })`** → `POST /agents/credentials/validate`. Validates an agent credential (`api_key` or `access_token`) and returns `{ valid, registrationId, expiresAt }`. Read-only — it never consumes or mutates the credential.

The module follows the existing per-module layout (`interfaces` / `serializers` / `fixtures` / spec), deserializing the API's snake_case payloads into camelCase. It's wired into the `WorkOS` client (`readonly agents`) and re-exported from `src/index.ts`.

## Notes

- **Hand-written, not generated by oagen.** These endpoints are gated behind the server-side `agent-auth-public-api` flag and excluded from the public OpenAPI spec, so they won't appear in codegen yet. When they're eventually published to the spec, oagen may want to regenerate this module.
- Calls will `404` against environments where `agent-auth-public-api` is not enabled — expected while the feature is flag-gated.

## Testing

- `jest src/agents` — 3 tests pass (registration read, valid credential, invalid credential)
- `tsc --noEmit`, `eslint`, and `prettier --check` all clean
- `index.barrel` spec passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Closes AUTH-6622

### DIFF
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -127,7 +127,12 @@ export class Agents {
     // identity. If it doesn't, treat the credential as invalid rather than
     // returning two conflicting registration identities in one result.
     if (remote.registrationId !== claims.registrationId) {
-      return { valid: false, registrationId: null, expiresAt: null, claims: null };
+      return {
+        valid: false,
+        registrationId: null,
+        expiresAt: null,
+        claims: null,
+      };
     }
 
     return { ...remote, claims };
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -126,10 +126,7 @@ export class Agents {
    * The audience defaults to the client ID; resource-scoped tokens carry the
    * resource as their audience and require it to be passed explicitly.
    */
-  private async verifyAccessTokenClaims(
-    credential: string,
-    audience?: string | string[],
-  ) {
+  private async verifyAccessTokenClaims(credential: string, audience?: string) {
     const { jwtVerify } = await getJose();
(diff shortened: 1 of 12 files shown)


m0tzy:
@greptileai

src/agents/agents.ts
```diff
@@ -0,0 +1,216 @@
...
+ */
+function hasRequiredAgentClaims(
+  payload: import('jose').JWTPayload,
+): payload is SerializedAgentAccessTokenClaims {
+  return (
+    typeof payload.iss === 'string' &&
+    (typeof payload.aud === 'string' || Array.isArray(payload.aud)) &&
+    typeof payload.sub === 'string' &&
+    typeof payload.jti === 'string' &&
+    typeof payload.organization_id === 'string' &&
```
imkesin, line 33: Interesting that this one is full `organization_id` even though we're often using the `org_id` shorthand.

imkesin (approved):

src/agents/agents.ts
m0tzy: Rats, this was not intentionally divergent

PR #1652, opened 2026-07-02
feat: Add linkClaimAttemptToExternalUser to Agents module

m0tzy:
## Description

Adds `workos.agents.linkClaimAttemptToExternalUser(options)` wrapping `PATCH /agents/claims/attempts`.

The serializer hardcodes `type: 'link_external_user'` so callers don't need to pass it:

```ts
const result = await workos.agents.linkClaimAttemptToExternalUser({
  claimAttemptToken: 'cla_tkn_...',
  user: { email: 'alice@example.com', externalId: 'ext_alice' },
  organizationId: 'org_...', // optional
});
// result: { id, status, userCode, organizations }
```

Types: `LinkClaimAttemptToExternalUserOptions`, `ClaimAttemptResponse`, `ClaimAttemptOrganization`

The endpoint uses `PATCH` because it modifies an existing claim attempt (the agent already created it) rather than creating one. Generated SDKs surface a generic `updateAttempts({ type: 'link_external_user', ... })`; this hand-written SDK keeps the ergonomic `linkClaimAttemptToExternalUser` wrapper.

## Documentation

```
[x] Yes
```

API reference docs updated in the companion monorepo PR (workos/workos#64535).

Closes AUTH-6647

Link to Devin session: https://app.devin.ai/sessions/4c6130f5dc744691a1c9fe409b253aac
Requested by: @m0tzy

### DIFF
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -64,7 +64,7 @@ export class Agents {
   async linkClaimAttemptToExternalUser(
     options: LinkClaimAttemptToExternalUserOptions,
   ): Promise<ClaimAttemptResponse> {
-    const { data } = await this.workos.post<SerializedClaimAttemptResponse>(
+    const { data } = await this.workos.patch<SerializedClaimAttemptResponse>(
       '/agents/claims/attempts',
       serializeLinkClaimAttemptToExternalUserOptions(options),
     );
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -3,7 +3,10 @@ import { WorkOS } from '../workos';
 import {
   AgentCredentialValidation,
   AgentRegistration,
+  ClaimAttemptResponse,
+  CreateClaimAttemptOptions,
   SerializedAgentAccessTokenClaims,
+  SerializedClaimAttemptResponse,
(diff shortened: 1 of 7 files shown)


src/agents/agents.ts
```diff
@@ -41,6 +46,32 @@ export class Agents {
...
+   * organizations, an explicit organization must be provided.
+   *
+   * @param options - Object containing the claim attempt token, user details, and optional organization ID.
+   * @returns {Promise<ClaimAttemptResponse>}
+   * @throws {BadRequestException} 400 - Invalid request, email mismatch, or wrong account.
+   * @throws {ForbiddenException} 403 - Claim denied or auth method disabled.
+   * @throws {ConflictException} 409 - Organization selection required, external ID conflict, or already claimed.
+   * @throws {GoneException} 410 - Claim or user code expired.
+   */
+  async linkClaimAttemptToExternalUser(
```
stanleyphu, line 64: I don't have a strong opinion here on the name, but just flagging that this is different from the other SDKs' generated method (`createAttempt`) so I'm wondering if we should rename this to something more similar or the same.

stanleyphu (approved):

PR #1659, opened 2026-07-16
feat: Add configurable automatic retries to the HTTP client

awolfden:
## Summary

Brings the Node SDK's automatic retry behavior to parity with the Kotlin, Go, Ruby, and Python SDKs.

Previously the Node SDK had retry machinery, but it was effectively dormant: retries only fired for two path prefixes (`/vault/*` and `/audit_logs/events`), never retried `429`, ignored `Retry-After`, added no idempotency safety for retried writes, and weren't configurable. Every other WorkOS SDK retries all requests on transient failures with backoff + jitter.

This PR makes retries apply to **all** requests and configurable, while keeping default behavior transparent to existing callers (default is still 3 attempts, backoff + jitter unchanged).

### What changed

- **Retries apply to every request**, not just the two allowlisted path prefixes. The per-verb `isPathRetryable(path)` branch is removed; all verbs go through `fetchRequestWithRetry`.
- **Retryable statuses** now include `429` and `503` (was `[408, 500, 502, 504]`, now `[408, 429, 500, 502, 503, 504]`). Transport errors (`TypeError`) are still retried.
- **`Retry-After` is honored, capped at 60 seconds.** When a retryable response carries a `Retry-After` header, the SDK sleeps for that duration instead of the computed backoff — capped at `MAXIMUM_RETRY_AFTER_TIME_IN_MILLISECONDS = 60_000` so a large server-provided delay can't hang callers. Supports both delay-seconds (strict RFC 9110 integer) and HTTP-date forms:

  ```ts
  await this.sleep(attempt, HttpClient.parseRetryAfter(retryAfterHeader));
  // parseRetryAfter('120') -> 120000; parseRetryAfter('3600') -> 60000 (capped);
  // parseRetryAfter('<http-date>') -> ms until date (capped); else null (fall back to backoff)
  ```

- **Idempotency-Key auto-injection** for retried `POST` requests. A generated key (`retry-<uuid>`) is attached once to `POST` requests that don't already have one, and reused across attempts so a retried write is not applied twice. A caller-provided key is always preserved. POST-only to match the Kotlin and Go SDKs and the API, which honors `Idempotency-Key` on create (POST) endpoints.
- **Backoff is now capped** (`MAXIMUM_SLEEP_TIME_IN_MILLISECONDS = 8_000`) to avoid unbounded delays, matching Kotlin's `maxDelayMs`.
- **Configurable `maxRetries`**, client-wide and per-request, `0` disables retries:

  ```ts
  const workos = new WorkOS({ apiKey, maxRetries: 5 });   // client-wide
  await workos.get('/path', { maxRetries: 0 });           // per-request override
  ```

  Wired through `WorkOSOptions`, the per-call option interfaces (`GetOptions`/`PostOptions`/`PutOptions`/`PatchOptions`), and `RequestOptions`. Per-request wins over client-wide, which falls back to the default of 3.

### Notes

- Because `408`/`429`/`5xx` are now retried by default, a couple of existing error-mapping tests that asserted a single terminal `500`/`429` response now construct their client with `maxRetries: 0` to isolate the mapping from retry behavior.

## Test plan

- Extended `src/common/net/fetch-client.spec.ts`: retries on non-allowlisted paths, `429`, `Retry-After` honored (delay-seconds and HTTP-date, 60s cap, unparseable fallback), idempotency key generated + reused on POST (and not on PUT/PATCH), caller key preserved, network-error retry, DELETE retry, `maxRetries: 0` disables, per-request override.
- `npm run lint`, `npm run typecheck`, `npm test` (888 passing), and `npm run build` all pass locally.
- A retries section was added to the README documenting the default behavior and `maxRetries` configuration.

Link to Devin session: https://app.devin.ai/sessions/45217aff3c0248e083c27dd69b391ca5
Requested by: @awolfden

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,29 @@ import { WorkOS } from '@workos-inc/node';
 const workos = new WorkOS('sk_1234');
 ```
 
+### Automatic retries
+
+The SDK automatically retries requests that fail with a transient error — a
+network error, a request timeout (408), a rate limit (429), or a server error
+(500, 502, 503, 504) — using exponential backoff with jitter. When the API
+responds with a `Retry-After` header, the SDK waits for that duration instead
+(capped at 60 seconds). Retried `POST` requests are automatically assigned an
+`Idempotency-Key` header (unless one is already set) so a retried write is not
+applied twice.
+
+The default is 3 retries. Configure it client-wide, or per request on the
+low-level HTTP methods:
+
+```ts
+// Client-wide: up to 5 retries per request
+const workos = new WorkOS('sk_1234', { maxRetries: 5 });
+
+// Per-request override: disable retries for this call only
+await workos.get('/organizations', { maxRetries: 0 });
+```
+
+Set `maxRetries: 0` to disable automatic retries entirely.
+
 ## Public Client Mode (Browser/Mobile/CLI)
 
 For apps that can't securely store secrets, initialize with just a client ID:
--- a/src/common/exceptions/parse-error.ts
+++ b/src/common/exceptions/parse-error.ts
@@ -7,21 +7,25 @@ export class ParseError extends Error implements RequestException {
   readonly rawBody: string;
   readonly rawStatus: number;
   readonly requestID: string;
+  readonly rawHeaders?: Headers;
 
   constructor({
     message,
     rawBody,
     rawStatus,
     requestID,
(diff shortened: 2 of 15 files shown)


gjtorikian (approved):
i pushed a few changes, looks good to me!

PR #1665, opened 2026-07-24
fix: `totp` on list-path `AuthenticationFactor` is now optional to match the API

jeffnv:
## Description

Fixes DAAP-3034 (Sentry API-2FR): `multiFactorAuth.listUserAuthFactors()` threw `TypeError: Cannot read properties of undefined (reading 'issuer')` for any user with a non-TOTP factor, failing the entire list call.

`GET /user_management/users/:id/auth_factors` returns factors of type `totp`, `sms`, or `generic_otp`, but only TOTP factors include a `totp` object (SMS factors carry `sms: { phone_number }` instead). The user-management `deserializeFactor` called `deserializeTotp(factor.totp)` unconditionally:

```diff
- totp: deserializeTotp(factor.totp),
+ ...(factor.sms ? { sms: deserializeSms(factor.sms) } : {}),
+ ...(factor.totp ? { totp: deserializeTotp(factor.totp) } : {}),
```

This matches the existing (correct) sibling deserializer in `src/multi-factor-auth/serializers/factor.serializer.ts`; the divergence between the two was the whole bug.

### Type fix (why the compiler never caught this)

The `AuthenticationFactor`/`AuthenticationFactorResponse` interfaces hard-coded `type: 'totp'` and a required `totp` object — true of the enroll endpoint, false of list. They now use the shared `FactorType` union (`'sms' | 'totp' | 'generic_otp'`, exported from the MFA module so there's a single source of truth) with optional `sms?` / `totp?`.

Judgment calls:
- **Semver**: widening `type` and making `totp` optional on the list path is technically breaking for TypeScript consumers who destructure `factor.totp` — but the type was lying; that code already crashes at runtime for exactly the users this affects. Shipping as a `fix` (patch) since it corrects the types to the API's actual, long-standing response contract.
- **`WithSecrets` (enroll path) stays narrow**: `EnrollAuthFactorOptions.type` is `'totp'` only, so `createUserAuthFactor` cannot return a non-TOTP factor; `deserializeFactorWithSecrets` keeps the unguarded call so a contract-violating response fails loudly rather than silently returning `undefined` (per review).
- **`sms` field**: now surfaced on the returned list object (previously dropped), matching the MFA sibling. Additive.

Test coverage: both `list-factors.json` fixtures now include `sms` and `generic_otp` factors, exercised through the `listUserAuthFactors` spec and a new `authentication-factor.serializer.spec.ts` — either would have caught the bug.

Note: the production consumer (WorkOS dashboard) is on SDK v8.10.0 (`userManagement.listAuthFactors()`); there is no 8.x maintenance branch, so it will pick this up on upgrade to v9+/v10 where the method is `multiFactorAuth.listUserAuthFactors()`.

## Documentation

```
[ ] No
```

Link to Devin session: https://app.devin.ai/sessions/7fa3a604bb6d415ab2989a8e59fcf403
Requested by: @jeffnv

### DIFF
--- a/src/user-management/fixtures/list-factors.json
+++ b/src/user-management/fixtures/list-factors.json
@@ -10,7 +10,27 @@
       "totp": {
         "issuer": "WorkOS",
         "user": "some_user"
-      }
+      },
+      "user_id": "user_01H5JQDV7R7ATEYZDEG0W5PRYS"
+    },
+    {
+      "object": "authentication_factor",
+      "id": "auth_factor_5678",
+      "created_at": "2022-03-15T20:39:19.892Z",
+      "updated_at": "2022-03-15T20:39:19.892Z",
+      "type": "sms",
+      "sms": {
+        "phone_number": "+15555555555"
+      },
+      "user_id": "user_01H5JQDV7R7ATEYZDEG0W5PRYS"
+    },
+    {
+      "object": "authentication_factor",
+      "id": "auth_factor_9012",
+      "created_at": "2022-03-15T20:39:19.892Z",
+      "updated_at": "2022-03-15T20:39:19.892Z",
+      "type": "generic_otp",
+      "user_id": "user_01H5JQDV7R7ATEYZDEG0W5PRYS"
     }
   ],
   "list_metadata": {
--- a/src/user-management/interfaces/authentication-factor.interface.ts
+++ b/src/user-management/interfaces/authentication-factor.interface.ts
@@ -1,17 +1,24 @@
 import {
+  Sms,
+  SmsResponse,
+} from '../../multi-factor-auth/interfaces/sms.interface';
+import {
   Totp,
   TotpResponse,
   TotpWithSecrets,
   TotpWithSecretsResponse,
 } from '../../multi-factor-auth/interfaces/totp.interface';
 
+export type AuthenticationFactorType = 'sms' | 'totp' | 'generic_otp';
+
 export interface AuthenticationFactor {
   object: 'authentication_factor';
   id: string;
   createdAt: string;
   updatedAt: string;
-  type: 'totp';
-  totp: Totp;
+  type: AuthenticationFactorType;
+  sms?: Sms;
+  totp?: Totp;
   userId: string;
 }
 
@@ -20,8 +27,9 @@ export interface AuthenticationFactorWithSecrets {
   id: string;
   createdAt: string;
   updatedAt: string;
-  type: 'totp';
-  totp: TotpWithSecrets;
+  type: AuthenticationFactorType;
+  sms?: Sms;
+  totp?: TotpWithSecrets;
   userId: string;
 }
 
@@ -30,8 +38,9 @@ export interface AuthenticationFactorResponse {
   id: string;
   created_at: string;
   updated_at: string;
-  type: 'totp';
-  totp: TotpResponse;
+  type: AuthenticationFactorType;
+  sms?: SmsResponse;
+  totp?: TotpResponse;
   user_id: string;
 }
 
@@ -40,7 +49,8 @@ export interface AuthenticationFactorWithSecretsResponse {
   id: string;
   created_at: string;
   updated_at: string;
-  type: 'totp';
-  totp: TotpWithSecretsResponse;
+  type: AuthenticationFactorType;
+  sms?: SmsResponse;
+  totp?: TotpWithSecretsResponse;
   user_id: string;
 }
--- a/src/user-management/interfaces/authentication-factor.interface.ts
+++ b/src/user-management/interfaces/authentication-factor.interface.ts
@@ -1,3 +1,4 @@
+import { FactorType } from '../../multi-factor-auth/interfaces/factor.interface';
 import {
   Sms,
   SmsResponse,
@@ -9,7 +10,7 @@ import {
   TotpWithSecretsResponse,
 } from '../../multi-factor-auth/interfaces/totp.interface';
 
-export type AuthenticationFactorType = 'sms' | 'totp' | 'generic_otp';
+export type AuthenticationFactorType = FactorType;
 
 export interface AuthenticationFactor {
   object: 'authentication_factor';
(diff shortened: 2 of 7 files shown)


src/user-management/interfaces/authentication-factor.interface.ts
```diff
@@ -20,8 +27,9 @@ export interface AuthenticationFactorWithSecrets {
   id: string;
   createdAt: string;
   updatedAt: string;
-  type: 'totp';
-  totp: TotpWithSecrets;
+  type: AuthenticationFactorType;
+  sms?: Sms;
+  totp?: TotpWithSecrets;
```
jeffnv: I'd keep the `WithSecrets` pair narrow — `type: 'totp'` with `totp: TotpWithSecrets` required — and only widen the non-secrets pair.

`EnrollAuthFactorOptions.type` is `'totp'` and nothing else (`enroll-auth-factor.interface.ts:3`), so `createUserAuthFactor` provably cannot return a non-TOTP factor. Widening here doesn't fix a bug; it just makes every enroll consumer null-check an impossible case before reading `.totp.secret` / `.qrCode` / `.uri`. Verified it's a real compile break: `error TS18048: 'f.totp' is possibly 'undefined'`.

The fair counter-argument: with the guard added to `deserializeFactorWithSecrets`, that function genuinely *can* now return an object without `totp`, so the optional type is honest about it. If you want the narrow type back, either drop the guard on the `WithSecrets` variant only — an enroll response missing its TOTP payload is a contract violation you'd rather throw on than silently return `undefined` for — or keep the guard and throw explicitly.

Same applies to `AuthenticationFactorWithSecretsResponse` at lines 52-54.

src/user-management/interfaces/authentication-factor.interface.ts
```diff
@@ -1,17 +1,24 @@
...
+  SmsResponse,
+} from '../../multi-factor-auth/interfaces/sms.interface';
 import {
   Totp,
   TotpResponse,
   TotpWithSecrets,
   TotpWithSecretsResponse,
 } from '../../multi-factor-auth/interfaces/totp.interface';
 
+export type AuthenticationFactorType = 'sms' | 'totp' | 'generic_otp';
```
jeffnv: This union already exists as `FactorType` in `multi-factor-auth/interfaces/factor.interface.ts:9` with identical members. Two independently-maintained copies of one union, across these exact two modules, is the divergence that caused this incident in the first place.

Worth exporting `FactorType` and reusing it here rather than declaring a second copy that can drift when a fifth factor type shows up.

src/user-management/fixtures/list-factors.json
```diff
@@ -10,7 +10,27 @@
       "totp": {
         "issuer": "WorkOS",
         "user": "some_user"
-      }
+      },
+      "user_id": "user_01H5JQDV7R7ATEYZDEG0W5PRYS"
```
jeffnv, line 14: This fixture is unreferenced — the only `list-factors` import in the tree is `multi-factor-auth.spec.ts:16` pulling `./fixtures/list-factors.json`, which resolves to the MFA copy.

Updating it reads as added coverage but contributes none. Worth deleting the file, or leaving a note on why it's kept.

jeffnv (commented):
Reviewed the full branch with the PR checked out locally.

## Verdict: correct fix, ship it after one type decision

The core change is right, and the test genuinely pins the bug — I verified that rather than trusting it. Reverting just the serializer hunk to the pre-fix form and re-running the spec reproduces the exact production failure:

```
TypeError: Cannot read properties of undefined (reading 'issuer')
  > 10 |     issuer: totp.issuer,
  at src/multi-factor-auth/serializers/totp.serializer.ts:10:18
Tests: 1 failed, 13 passed
```

Restored, it passes. Also confirmed locally: `tsc --noEmit` clean, 902 tests pass across 48 suites, all 7 CI checks green.

One thing the diff obscures and that's worth stating for future readers: `listUserAuthFactors` lives in the MFA module but imports the *user-management* deserializer (`multi-factor-auth.ts:36`, aliased `deserializeUMFactor`). So the test added to `multi-factor-auth.spec.ts` really does cover the fixed code path — it isn't exercising the sibling by mistake.

## One finding I'd act on

**The `WithSecrets` widening is a consumer break with no correctness benefit.** Details inline, but in short: `EnrollAuthFactorOptions.type` is `'totp'` and nothing else (`enroll-auth-factor.interface.ts:3`), so `createUserAuthFactor` cannot return a non-TOTP factor. Making `totp` optional there forces every enroll consumer to null-check an impossible case.

Compiling a representative consumer against the new types:

```
probe.ts(3,69): error TS18048: 'f.totp' is possibly 'undefined'.   // enroll: f.totp.qrCode
probe.ts(5,60): error TS18048: 'f.totp' is possibly 'undefined'.   // list:   f.totp.issuer
```

The list break is justified — that code is already crashing at runtime for SMS users, so the compile error surfaces a real bug. The enroll break is pure cost, and it hits the consumers reading `.totp.secret` / `.qrCode` / `.uri`, which is most of them.

The PR description draws this distinction correctly ("true of the enroll endpoint, false of list") and then widens both anyway for consistency — trading an accurate type for a uniform one.

## Two nits

Both inline: a duplicated `'sms' | 'totp' | 'generic_otp'` union, and a dead fixture file.

## Operational note (not a defect — the PR discloses it)

This doesn't reach the affected customer. Production is on SDK 8.10.0 via `userManagement.listAuthFactors()`, there is no 8.x maintenance branch, and the method has since moved to `multiFactorAuth.listUserAuthFactors()`. The dashboard only picks this up on a v9/v10 major upgrade.

So the `workos/workos`-side change is still the one that actually unblocks them, and the three unguarded call sites in `user-authentication-factors.resolver.ts` (`:87` list, `:272` enroll, `:337` delete) remain live regardless of what happens here.

src/user-management/serializers/authentication-factor.serializer.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+import { AuthenticationFactorResponse } from '../interfaces/authentication-factor.interface';
+import listFactorsFixture from '../fixtures/list-factors.json';
+
+describe('deserializeFactor', () => {
+  it('deserializes totp, sms, and generic_otp factors', () => {
+    const factors = listFactorsFixture.data.map((factor) =>
+      deserializeFactor(factor as AuthenticationFactorResponse),
+    );
+
+    expect(factors).toEqual([
```
gjtorikian: I'd suggest `toStrictEqual`

gjtorikian (approved):

PR #1683, opened 2026-08-21
feat!: Add flag targets and flag target types resources

Deborah-Digges:
## Description

Adds the oagen-generated SDK surface for the new flag target endpoints:

- `workos.flagTargets` — `listFlagTargets`, `createFlagTarget`, `getFlagTarget`, `deleteFlagTarget` (`/flag_targets`)
- `workos.flagTargetTypes` — `listFlagTargetTypes`, `createFlagTargetType`, `getFlagTargetType` (`/flag_target_types`)

All resource classes, interfaces, serializers, tests, and fixtures under `src/flag-targets/` and `src/flag-target-types/` are generated by oagen from the OpenAPI spec (no hand-written wrappers, per the ticket). Hand-edited files are limited to wiring the accessors in `src/workos.ts` and the interface exports in `src/index.ts`.

**Naming note (potentially breaking, type-only):** the runtime client's poll-response interface `FlagTarget { id; enabled }` is renamed to `FlagPollTarget`, freeing the `FlagTarget` name for the API resource model:

```diff
-export interface FlagTarget { id: string; enabled: boolean }
+export interface FlagPollTarget { id: string; enabled: boolean }
+// FlagTarget is now the /flag_targets resource:
+export interface FlagTarget { object: 'flag_target'; id; flagId; flagSlug; environmentId; targetType; targetId; valueType; value; createdAt; updatedAt }
```

Without the rename, the emitter aliases the resource's return type to the existing (incompatible) poll shape. Only type-level imports of `FlagTarget` from the poll response are affected; runtime behavior is unchanged.

The endpoints are still `@ApiExcludeController()` in the monorepo; making them public in the spec/docs ships separately once SDKs are released.

Fixes AUTH-6891

## Documentation

```
[ ] Yes
```

API reference docs for these endpoints are generated from the monorepo spec and will ship with the PR that removes `@ApiExcludeController()`.

Link to Devin session: https://app.devin.ai/sessions/fdf25eeb8122400891ee756d079bfbeb
Requested by: @Deborah-Digges


src/feature-flags/interfaces/flag-poll-response.interface.ts
```diff
@@ -1,4 +1,4 @@
-export interface FlagTarget {
+export interface FlagPollTarget {
```
nicknisi, line 1: question: is renaming this interface a breaking change?

nicknisi: IMO we should avoid a major version if at all possible, but would love to hear @Deborah-Digges' thoughts!

Deborah-Digges: @nicknisi it's hard to get away from this rename because `FlagTarget` is the name of the new resource (that is currently being used by this internal implementation class, for which I think `FlagPollTarget` is a better name) 💭 

If we want to do a phased rollout, to be careful, we could potentially do something like this:

- Leave the poll interface named `FlagTarget`, add export type `FlagPollTarget = FlagTarget;` plus a `@deprecated` note on `FlagTarget` so people can migrate ahead of time
- Drop the two new export * lines from the root barrels this release. `workos.flagTargets.*` still returns fully-typed results by inference — consumers just can't import type { FlagTarget } for the resource yet. 
- Next major: delete the poll FlagTarget, keep FlagPollTarget, add the star exports.

It would still require a major release (deferred), if that is preferable, although it would be cleaner to do the major release now if that is ok with the team 🙏🏻

nicknisi: if it's unavoidable, that's fine. Just want to make sure we follow strict semver. Please change the PR title to include `feat!:` so release-please properly tags it as a major/breaking change.

```
feat!: Add flag targets and flag target types resources
```

Deborah-Digges:
Closing this because there is a different paved path to create SDK PRs

PR #1711, opened 2026-09-21
fix: keep request timeouts active while reading response bodies

### LINKED ISSUE #1679: FetchHttpClient: request timeout stops covering the response body once headers arrive
## Problem

In `src/common/net/fetch-client.ts` (~lines 212–214), the per-request timeout's `clearTimeout` fires as soon as `fetch` resolves — i.e., when **headers** arrive. Reading the response body after that has no deadline, so a server that sends headers promptly but streams the body slowly (or stalls mid-body) can hang the SDK call indefinitely despite a configured `options.timeout`.

## Suggested fix

Keep the per-request `AbortController` (created at ~lines 192–195) armed until the body is fully consumed, or apply a separate body-read deadline, instead of clearing the timer on header receipt.

## Related minor nit (same file)

`toJSON()` (~lines 431–435) returns `null` without draining the body when the content-type isn't JSON, leaving the response unconsumed. Undici destroys that socket (safe), but it's wasted connection churn — draining or explicitly cancelling the body would be cleaner. Could be fixed in the same pass.

CodingCossack:
Requests could hang indefinitely when a server returned headers but stalled the response body.

Keep the original per-attempt timeout active through body consumption. Preserve the original response object, prevent competing readers from disarming another reader’s deadline, and leave unrelated body-read errors unchanged.

Timeouts while reading error responses remain subject to the existing retry policy, preserving Retry-After and POST idempotency keys. Timeouts while reading successful responses are reported as 408 errors but are not automatically retried.

Validated with deterministic regressions and real HTTP tests using native Fetch and node-fetch 2.7.0. The full suite passes with 1,061 tests, alongside typecheck, lint, formatting and build checks.

Fixes #1679.


gjtorikian:
thanks !

CodingCossack:
glad i could help. all the best :)

